PR #6, opened 2018-07-24
Warn when a app / pod ID is longer than 54 chars

bamarni:
Such service IDs might not play well with Mesos DNS due to SRV records length limitation.

https://mesosphere.github.io/marathon/docs/application-basics.html#service-name-length-limitations
https://jira.mesosphere.com/browse/DCOS_OSS-3722


leehambley:
LGTM. (I'm not one of the two reviewers, so I can't sign this off, but for such a small change I don't know that it's necessary)

PR #13, opened 2018-08-20
Fix incorrect plugins breaking package list

br-lewis:
Change how subcommands are found to filter out any subcommands that don't have a `package.json` file. This will prevent invalid subcommand structures from causing the CLI to crash


bamarni:

PR #16, opened 2018-08-23
Remove configured clusters before integration tests

bamarni:
This makes sure there is no leftover clusters from other builds, as these tests expect a single cluster to be configured.


bamarni:

PR #19, opened 2018-08-23
Fixes table printer when rendering empty data

rgo3:
With this commit, the command
`dcos job show runs <job-id> --run-id <run-id>` works correctly.
It was previously failing due to a type error.

https://jira.mesosphere.com/browse/DCOS_OSS-3974


rgo3:

PR #26, opened 2018-09-06
Added Go subcommand: `job run`.

rgo3:
This subcommand is currently not user-facing, it will be once
all job subcommands will be refactored using Golang.

https://jira.mesosphere.com/browse/DCOS_OSS-4063

### DIFF
--- a/pkg/cmd/job/job_run.go
+++ b/pkg/cmd/job/job_run.go
@@ -1,0 +1,25 @@
+package job
+
+import (
+	"github.com/dcos/dcos-cli/api"
+	"github.com/spf13/cobra"
+)
+
+// newCmdClusterRun runs a given job right now.
+func newCmdJobRun(ctx api.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "run",
+		Short: "run a job now",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			client, err := metronomeClient(ctx)
+			if err != nil {
+				return err
+			}
+
+			_, err = client.RunJob(args[0])
+			return err
+		},
+	}
+	return cmd
+}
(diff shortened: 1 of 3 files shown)


pkg/cmd/job/job_run.go
```diff
@@ -0,0 +1,30 @@
...
+				return err
+			}
+
+			runID := args[0]
+			err = client.RunJob(runID)
+			if err != nil {
+				return err
+			}
+
+			return nil
```
bamarni: This could simply be `return client.RunJob(args[0])`

pkg/metronome/client.go
```diff
@@ -76,3 +78,26 @@ func (c *Client) Jobs(opts ...JobsOption) ([]Job, error) {
 
 	return jobs, err
 }
+
+// RunJob triggers a run of the job with a given runID right now.
+func (c *Client) RunJob(runID string) error {
```
bamarni: The APi says it returns a response body : 
```
{
  "completedAt": null,
  "createdAt": "2016-07-15T13:02:59.735+0000",
  "id": "20160715130259A34HX",
  "jobId": "prod",
  "status": "STARTING",
  "tasks": []
}
```
We should also return it from this method.

pkg/metronome/client.go
```diff
@@ -76,3 +78,26 @@ func (c *Client) Jobs(opts ...JobsOption) ([]Job, error) {
...
+	resp, err := c.http.Post("/v1/jobs/"+runID+"/runs", "application/json", nil)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 201 {
```
bamarni: In general it is more readable to return early in such cases (as it avoids too many nested block levels) : 
``` go 
	if resp.StatusCode == 201 {
		return nil
	}
	if resp.StatusCode == 404 {
		return fmt.Errorf("job %s does not exist", runID)
	}
 	var apiError *dcos.Error
	if err := json.NewDecoder(resp.Body).Decode(&apiError); err != nil {
		return err
	}
	return apiError
```

A switch statement might even be more suited than the above here.

PR #28, opened 2018-09-06
Fixed time format

rgo3:
Introduced our own time constant because `time.RFC3339` was not matching the time format which is returned by the metronome API.

### DIFF
--- a/pkg/metronome/types.go
+++ b/pkg/metronome/types.go
@@ -132,28 +132,28 @@ func (j *Job) Status() string {
 }
 
 // LastRunStatus returns the status of the last run of this job.
-func (j *Job) LastRunStatus() (string, error) {
+func (j *Job) LastRunStatus() string {
 	if j.HistorySummary.LastSuccessAt == "" && j.HistorySummary.LastFailureAt == "" {
-		return "N/A", nil
+		return "N/A"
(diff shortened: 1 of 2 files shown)


pkg/metronome/types.go
```diff
@@ -139,12 +141,12 @@ func (j *Job) LastRunStatus() (string, error) {
 		return "Failure", nil
 	}
 
-	lastSuccess, err := time.Parse(time.RFC3339, j.HistorySummary.LastSuccessAt)
+	lastSuccess, err := time.Parse(apiTimeFormat, j.HistorySummary.LastSuccessAt)
 	if err != nil {
 		return "", err
```
bamarni: Not directly related to this change, but I'm wondering if returning `N/A` here wouldn't be more appropriate, and skipping the error from the function signature.

Otherwise if we keep the error we might also want not to have the `LastRunStatus` method on the Job struct directly, as it is basically a representation of the API response, it shouldn't hold too much logic.

A reason why I mention this is that, with the current implementation and assuming there is a wrong formatting on one of these values returned from the API, using the list command with the `--json` or `--quiet` flags would succeed while displaying the table would fail. And I'd expect them to be consistent in that aspect, it's just different representations of the same data set.

PR #30, opened 2018-09-07
WIP: Added `job add` and `job update` subcommands.

rgo3:


pkg/metronome/client.go
```diff
@@ -79,6 +83,57 @@ func (c *Client) Jobs(opts ...JobsOption) ([]Job, error) {
...
+
+	if err = json.NewDecoder(resp.Body).Decode(&job); err != nil {
+		return nil, err
+	}
+
+	return &job, nil
+}
+
+// Add creates a job
+func (c *Client) Add(filename string) (*Job, error) {
```
bamarni: I'd rather make this taking a `*Job` and parsing the filename in the command, to have a clear separation between the CLI and the client.

pkg/cmd/job/job_update.go
```diff
@@ -0,0 +1,25 @@
...
+		Use:   "update",
+		Short: "update a job",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			client, err := metronomeClient(ctx)
+			if err != nil {
+				return err
+			}
+
+			_, err = client.Update(args[0])
```
bamarni: As a side note, it could be nice to also accepts reading from stdin instead of only having path to files.

pkg/metronome/client.go
```diff
@@ -79,6 +83,57 @@ func (c *Client) Jobs(opts ...JobsOption) ([]Job, error) {
 	return jobs, err
 }
 
+func (c *Client) addOrUpdate(filename string, add bool) (*Job, error) {
```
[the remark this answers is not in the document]
kensipe: This function has way too high cyclomatic complexity for what it is doing... It is 39 lines long with lots of line separators... if removed it is effectively 33 lines of code with 7 returns in it.

pkg/cmd/job/job_add.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// newCmdClusterRun runs a given job right now.
+func newCmdJobAdd(ctx api.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add",
+		Short: "add a job",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			client, err := metronomeClient(ctx)
```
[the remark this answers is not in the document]
bamarni: @kensipe : interesting, I hadn't thought about that. The CLI command is called `job` and not `metronome` so I can imagine there were some thoughts behind this decision.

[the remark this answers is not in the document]
rgo3: The thoughts behind this were that we wanted to make a metronome package in the pkg directory that could be imported by other projects. So I would keep the package name as is.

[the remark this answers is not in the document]
bamarni: Calling this helper function `jobClient` sounds ok, however the thing with `job` is that it's not a great name for a go package, from https://blog.golang.org/package-names : 

> Don't steal good names from the user. Avoid giving a package a name that is commonly used in client code. For example, the buffered I/O package is called bufio, not buf, since buf is a good variable name for a buffer.

And `job` is probably the most common variable name package callers would want to use.

pkg/metronome/client.go
bamarni: I think this should take a Job struct directly as argument, and the caller is responsible for parsing the stream. It'd act as a basic sanity check, at least when there is something obviously broken in the JSON we don't have to bother with calling the server just to get a parse error in return. It'd also make the package API more readable for users, as having well-defined arguments is easier to understand than simply streams.

pkg/cmd/job/job_add.go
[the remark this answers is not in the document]
kensipe: I would *NOT* keep the package as named metronome... we have a job service.  It is in the UI as job... it is a job.  metronome is the implementation of jobs and there is a talk about replacing it at some pt with another thing but that thing maintains the same v1 API of metronome.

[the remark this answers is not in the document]
bamarni: @kensipe : I agree that we don't call it metronome in the UI, even in the CLI the command is called `job` so we are fine on that aspect. However as a developer I think it makes sense to have a metronome client for the metronome project, the naming is just not exposed to the end-user.

In the future if we create another project which is compatible with the metronome API we could rename the package, or even keep it as is. As an analogy a mysql client can also work for MariaDB.

If you really really think it shouldn't be called `metronome` we should come-up with a good enough name, `job` is not an option for a Go package (see the link above), maybe `jobclient`.

pkg/cmd/job/job_add.go
```diff
@@ -0,0 +1,72 @@
...
+				reader = os.Stdin
+				break
+			case 1:
+				reader, err = os.Open(args[0])
+				if err != nil {
+					return err
+				}
+				break
+			default:
+				return fmt.Errorf("input must be from stdin or file")
```
bamarni: This could use the Args field, eg. https://github.com/dcos/dcos-cli/blob/96375b35f1d379f092f23a5e4f677a0b73200c1e/pkg/cmd/auth/auth_listproviders.go#L19

pkg/cmd/job/job_add.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/dcos/dcos-cli/api"
+	"github.com/dcos/dcos-core-cli/pkg/metronome"
+	"github.com/spf13/cobra"
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+// newCmdClusterAdd creates a new job.
+func newCmdJobAdd(ctx api.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add",
```
bamarni: When there is an arg we're adding it here, see https://github.com/dcos/dcos-cli/blob/96375b35f1d379f092f23a5e4f677a0b73200c1e/pkg/cmd/auth/auth_listproviders.go#L17

pkg/cmd/job/job_add.go
```diff
@@ -0,0 +1,72 @@
...
+			var reader io.Reader
+			switch len(args) {
+			case 0:
+				if terminal.IsTerminal(int(os.Stdin.Fd())) {
+					return fmt.Errorf("input from the terminal is not accepted")
+				}
+				reader = os.Stdin
+				break
+			case 1:
+				reader, err = os.Open(args[0])
```
bamarni: We have an abstraction for filesystem operations (to ease testing), it relies on the afero library and is in `ctx.Fs()`.

pkg/cmd/job/job_add.go
```diff
@@ -0,0 +1,72 @@
...
+			client, err := metronomeClient(ctx)
+			if err != nil {
+				return err
+			}
+
+			// Handling input from file or stdin
+			var reader io.Reader
+			switch len(args) {
+			case 0:
+				if terminal.IsTerminal(int(os.Stdin.Fd())) {
```
bamarni: It should rely on `ctx.Input()` (would need a conversion from io.Reader to os.File).

pkg/cmd/job/job_update.go
```diff
@@ -0,0 +1,57 @@
...
+	cmd := &cobra.Command{
+		Use:   "update",
+		Short: "update a job",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			client, err := metronomeClient(ctx)
+			if err != nil {
+				return err
+			}
+
+			// Handling input from file or stdin
```
bamarni: Does it make sense to have a helper function for this? The same way that we have `metronomeClient()`, eg. `inputReader()`

PR #32, opened 2018-09-11
Added `task attach` subcommand

klueska:


python/lib/dcos/dcos/mesos.py
```diff
@@ -1156,7 +1165,77 @@ def __init__(self, task_id, cmd=None, args=None,
         # exiting.
         self.exception = None
 
-    def run(self):
+        # Allow an exit sequence to be used to break the CLIs attachment to
+        # the remote task. Depending on the call, this may be disabled, or
+        # the exit sequence to be used may be overwritten.
+        self.supports_exit_sequence = True
+        self.exit_sequence = b'\x10\x11' # Ctrl-p, Ctrl-q
```
ArmandGrillet: There should be at least two spaces before an inline comment (E261).

python/lib/dcos/dcos/mesos.py
```diff
@@ -1192,7 +1271,8 @@ def run(self):
         try:
             if self.interactive:
                 tty.setraw(fd, when=termios.TCSANOW)
-                self._window_resize(signal.SIGWINCH, None)
+                self._window_resize(signal.SIGWINCH, dimensions=[0,0])
```
ArmandGrillet: missing whitespace after ',' (E231).

python/lib/dcos/dcos/mesos.py
```diff
@@ -493,17 +493,33 @@ def tasks(self, fltr=None, completed=False, all_=False):
...
+                        for task in framework['tasks']:
+                            if 'id' in task:
+                                if task['id'].startswith(task_id):
+                                    candidates.append(task)
+
+            if len(candidates) == 1:
+                return candidates[0]
+
+            raise DCOSException(
+                "More than one task matching '{}' found: {}"
```
bamarni: This message would also be printed in case there is no match.

python/lib/dcos/dcos/mesos.py
```diff
@@ -493,17 +493,33 @@ def tasks(self, fltr=None, completed=False, all_=False):
...
-            task = task_obj.dict()
+        def _get_task(task_id):
+            candidates = []
+            if 'frameworks' in self.state():
+                for framework in self.state()['frameworks']:
+                    if 'tasks' in framework:
+                        for task in framework['tasks']:
+                            if 'id' in task:
+                                if task['id'].startswith(task_id):
+                                    candidates.append(task)
```
bamarni: There are a lot of indentation levels, if conditions before loops could be removed by using the `dict.get('key', default)` syntax:
``` python
for framework in self.state().get('frameworks', []):
    for task in framework.get('tasks', []):
        # ...
```

python/lib/dcos/dcos/mesos.py
```diff
@@ -1646,7 +1647,7 @@ def _heartbeat_thread(self):
             self.input_queue.put(self.encoder.encode(message))
             time.sleep(interval)
 
-    def _window_resize(self, signum, frame):
+    def _window_resize(self, signum=None, frame=None, dimensions=None):
         """Signal handler for SIGWINCH.
 
         Generates a message with the current demensions of the
```
bamarni: typo: `dimensions`

python/lib/dcos/dcos/mesos.py
```diff
@@ -1271,7 +1271,8 @@ def _run(self):
         try:
             if self.interactive:
                 tty.setraw(fd, when=termios.TCSANOW)
-                self._window_resize(signal.SIGWINCH, None)
+                self._window_resize(signal.SIGWINCH, dimensions=[0, 0])
+                self._window_resize(signal.SIGWINCH)
```
bamarni: Regarding the limitation in your commit message (`you are not able to resize the screen
and have it react properly`), would it make sense to catch the SIGWINCH signal and then call `_window_resize`? Maybe as a possible future improvement.

klueska: Calling this function sends a message to the TTY on the remote machine that a local resize has happened. When we first attach I want to send two messages: one setting the remote sides size to 0, followed by a second message setting it to the real size of the local terminal. Doing this causes the remote end to handle its SIGWINCH signal, thus redrawing the screen, and having that reflected locally.

Im not sure exactly what you are suggesting here, because there will be no local SIGWINCH signal to catch initially. This is just an artificial resize triggered on the remote end to force it to redraw its screen.

Also note, this is the same technique used by the docker CLI to force a refresh on attach.

klueska: Calling this function sends a message to the TTY on the remote machine that a local resize has happened. When we first attach I want to send two messages: one setting the remote sides size to 0, followed by a second message setting it to the real size of the local terminal. Doing this causes the remote end to handle its SIGWINCH signal, thus redrawing the screen, and having that reflected locally.

Im not sure exactly what you are suggesting here, because there will be no local SIGWINCH signal to catch initially. This is just an artificial resize triggered on the remote end to force it to redraw its screen.

Also note, this is the same technique used by the docker CLI to force a refresh on attach.

klueska: I should add a comment in line about this regardless a d link to the conversation about how/where it was added fir docker.

bamarni: Ah I see, I missed the line just below (`signal.signal(signal.SIGWINCH, self._window_resize)`), that's what I meant. Thx for the info.

bamarni:

klueska:
run linux integration tests

klueska:

ArmandGrillet:

klueska:

klueska:

ArmandGrillet:
Commit related to the failure we see: https://github.com/dcos/dcos-cli/commit/207d71f04450fbacdac2bbfc00a39d80a44043bf

ArmandGrillet:
To be more precise, `test_exec_match_id_pattern` is failing as currently we match any substring within the ID, not just the prefix of the ID, and this PR reverts this feature. Updating the test to:
```
def test_exec_match_id_pattern():
    assert_command(['dcos', 'task', 'exec', 'test-app1', 'true'])
    assert_command(['dcos', 'task', 'exec', 'test-app2', 'true'])
    code, _, _ = exec_command(['dcos', 'task', 'exec', 'test-app', 'true'])
    assert code != 0
```
Works but removes the main purpose of the test (we create apps called `test-app1` and `test-app2` before running the test).

klueska:
How have any PRs been passing this test? There are no changes in this PR that modify this behaviour.

klueska:
I see what you are saying. If I accidentally reverted some feature, we should undo that revert, not change the test.

PR #34, opened 2018-09-17
Added Go subcommand: `job kill`.

rgo3:
https://jira.mesosphere.com/browse/DCOS_OSS-4133

### DIFF
--- a/pkg/cmd/job/job.go
+++ b/pkg/cmd/job/job.go
@@ -24,6 +24,7 @@ func NewCommand(ctx api.Context) *cobra.Command {
 
 	cmd.AddCommand(
 		newCmdJobAdd(ctx),
+		newCmdJobKill(ctx),
 		newCmdJobList(ctx),
 		newCmdJobUpdate(ctx),
 		newCmdJobRemove(ctx),
--- a/pkg/cmd/job/job.go
+++ b/pkg/cmd/job/job.go
@@ -45,16 +45,19 @@ func metronomeClient(ctx api.Context) (*metronome.Client, error) {
 		baseURL = cluster.URL() + "/service/metronome"
 	}
 
-	return metronome.NewClient(httpclient.New(
-		baseURL,
-		httpclient.Logger(ctx.Logger()),
-		httpclient.ACSToken(cluster.ACSToken()),
-		httpclient.Timeout(cluster.Timeout()),
-		httpclient.TLS(&tls.Config{
-			InsecureSkipVerify: cluster.TLS().Insecure,
-			RootCAs:            cluster.TLS().RootCAs,
-		}),
-	)), nil
+	return metronome.NewClient(
+		httpclient.New(
+			baseURL,
+			httpclient.Logger(ctx.Logger()),
+			httpclient.ACSToken(cluster.ACSToken()),
+			httpclient.Timeout(cluster.Timeout()),
+			httpclient.TLS(&tls.Config{
+				InsecureSkipVerify: cluster.TLS().Insecure,
+				RootCAs:            cluster.TLS().RootCAs,
+			}),
+		),
+		ctx.Logger(),
+	), nil
 
 }
 
--- a/pkg/cmd/job/job_kill.go
+++ b/pkg/cmd/job/job_kill.go
@@ -29,25 +29,19 @@ func newCmdJobKill(ctx api.Context) *cobra.Command {
 				return err
 			}
 
-			var deadpool []string
-			if len(args) == 1 && all == true {
-				runs, err := client.Runs(args[0])
-				if err != nil {
-					return err
-				}
-				for _, run := range runs {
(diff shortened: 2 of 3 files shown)


pkg/cmd/job/job_kill.go
```diff
@@ -0,0 +1,61 @@
...
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			client, err := metronomeClient(ctx)
+			if err != nil {
+				return err
+			}
+
+			var deadpool []string
+			if len(args) == 1 && all == true {
+				runs, err := client.Runs(args[0])
```
bamarni: How the runs from here are different from the ones in `Job.ActiveRuns`?

rgo3: We could get them through that field too but I figured I would use the extra endpoint the API provides for it.

pkg/cmd/job/job_kill.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/dcos/dcos-cli/api"
+	"github.com/spf13/cobra"
+)
+
+// newCmdJobKill kills a run of a job
+func newCmdJobKill(ctx api.Context) *cobra.Command {
+	var all bool
+	cmd := &cobra.Command{
+		Use:   "kill <job-id> [OPTIONS]",
```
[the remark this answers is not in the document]
bamarni: sounds good

pkg/cmd/job/job_kill.go
bamarni: ok, I was just curious if there were any difference between the 2

pkg/cmd/job/job_kill.go
```diff
@@ -0,0 +1,58 @@
+package job
+
+import (
+	"fmt"
+
+	"github.com/dcos/dcos-cli/api"
+	"github.com/spf13/cobra"
+)
+
+// newCmdJobKill kills a run of a job
```
ArmandGrillet: Missing period.

pkg/cmd/job/job_kill.go
```diff
@@ -0,0 +1,58 @@
...
+				if err != nil {
+					return err
+				}
+				ctx.Logger().Infof("Run '%s' for job '%s' killed.", dead, args[0])
+			}
+
+			return nil
+		},
+	}
+	cmd.Flags().BoolVar(&all, "all", false, "kill all active runs of this job")
```
ArmandGrillet: Would add a `the`.

pkg/cmd/job/job_kill.go
```diff
@@ -0,0 +1,58 @@
...
+			} else if len(args) == 2 {
+				deadpool = append(deadpool, args[1])
+			}
+
+			for _, dead := range deadpool {
+				err = client.Kill(args[0], dead)
+				if err != nil {
+					return err
+				}
+				ctx.Logger().Infof("Run '%s' for job '%s' killed.", dead, args[0])
```
ArmandGrillet: s/for/of

pkg/cmd/job/job.go
```diff
@@ -25,6 +25,7 @@ func NewCommand(ctx api.Context) *cobra.Command {
 	cmd.AddCommand(
 		newCmdJobAdd(ctx),
 		newCmdJobList(ctx),
+		newCmdJobKill(ctx),
```
ArmandGrillet: Please move it it before `newCmdJobList`.

pkg/cmd/job/job.go
```diff
@@ -25,6 +25,7 @@ func NewCommand(ctx api.Context) *cobra.Command {
...
 		newCmdJobUpdate(ctx),
```
ArmandGrillet: Please move it after `newCmdJobRun`.

rgo3: I have done that on the `job schedule` command which you will see on that PR for that.

pkg/cmd/job/job_kill.go
```diff
@@ -0,0 +1,58 @@
...
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
+			client, err := metronomeClient(ctx)
+			if err != nil {
+				return err
+			}
+
+			var deadpool []string
+			if len(args) == 1 && all == true {
```
bamarni: Checking all is unnecessary here as there is a pre-check. I'd fine it more clear if there were an early exit path for a specific run-id and a single loop for the `--all` flag, as reading through the shared deadpool slice and the 2 loops is harder to grasp , like this for example : 
``` go
if len(args) == 2 {
	return client.Kill(args[0], args[1])
}

runs, err := client.Runs(args[0])
if err != nil {
	return err
}
for _, run := range runs {
	err = client.Kill(args[0], run.ID)
	if err != nil {
		return err
	}
}
```

PR #35, opened 2018-09-18
remove the pinning to python3.5

br-lewis:
Removes hard dependency on python3.5 to allow building of the python CLI on 3.7

### DIFF
--- a/python/bin/common.sh
+++ b/python/bin/common.sh
@@ -21,7 +21,7 @@ if [ "$(uname)" = "Windows_NT" ]; then
 else
   BIN=bin
   EXE=
-  : ${PYTHON:=python3.5${EXE}}
+  : ${PYTHON:=python3${EXE}}
 fi
 
 BASEDIR=$( cd "$(dirname $(dirname "${0}"))" > /dev/null; pwd -P )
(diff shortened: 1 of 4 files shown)


python/bin/common.sh
```diff
@@ -21,7 +21,7 @@ if [ "$(uname)" = "Windows_NT" ]; then
 else
```
ArmandGrillet: Let's add a check for `DCOS_EXPERIMENTAL` and then use `: ${PYTHON:=python3${EXE}}`, otherwise let's keep `: ${PYTHON:=python3.5${EXE}}`.

bamarni: The idea with `DCOS_EXPERIMENTAL` is to control the minor version check, whether or not it should be a strict comparison or a "less than". This `python3.5` is only the default here, one can already overwrite it by exporting `$PYTHON`

bamarni:
Windows build is failing : `bin/env.sh: option not availible on this NT BASH release`

ArmandGrillet:
@mesosphere-ci run continuous-integration/jenkins/pr-merge

ArmandGrillet:
@mesosphere-ci run continuous-integration/jenkins/pr-merge

ArmandGrillet:
@mesosphere-ci run continuous-integration/jenkins/pr-merge

ArmandGrillet:

PR #39, opened 2018-09-21
Cleanup 'ls' integration tests

klueska:
*  Cleaned up the logic for how we assert output in 'ls' integration tests.
    
    The number of output lines in the 'ls' tests depends on the number of
    tasks we are 'ls-ing' at the same time. The logic in these tests now
    reflects that so that if new test apps are added in the future, the code
    in these tests does not have to change.

*  Updated 'ls' integration tests to strip trailing newline from output.
    
    The number of lines being counted was unintuitive because of the
    trailing newline from the output. We new strip this trailing newline so
    that our assertion in the number of lines returned makes more sense.

*  Updated 'ls' integration tests to only look for 'stderr' and 'stdout'.
    
    Depending on the security mode of the cluster there might also be
    '.ssl*' and '.secret*' files in the sandbox directory in addition to the
    standard 'stderr' and 'stdout' files. The 'stderr' and 'stdout' files
    will always be present in all cases.
    
    Since we are only trying to test to make sure that ls works, it makes
    sense to only make sure that some subset of the files that we can
    guarantee to always be there will be present.

### DIFF
--- a/python/lib/dcoscli/tests/integrations/test_task.py
+++ b/python/lib/dcoscli/tests/integrations/test_task.py
@@ -202,13 +202,12 @@ def test_ls_no_params():
 
     ls_line = '.*stderr.*stdout.*'
     lines = stdout.decode('utf-8').rstrip().split('\n')
-    assert len(lines) == 6
-    assert re.match('===>.*<===', lines[0])
-    assert re.match(ls_line, lines[1])
-    assert re.match('===>.*<===', lines[2])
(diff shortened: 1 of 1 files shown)


klueska:

ArmandGrillet:

rgo3:

klueska:

ArmandGrillet:

PR #41, opened 2018-09-25
Updated node summary JSON to omit metric if unavailable.

klueska:
The new telegraf based metrics service will sometimes omit metrics for
'cpu.total' if there is no load on the CPU for a node. The CLI needs to
be able to handle this case gracefully.

### DIFF
--- a/python/lib/dcoscli/dcoscli/metrics.py
+++ b/python/lib/dcoscli/dcoscli/metrics.py
@@ -95,13 +95,21 @@ def _node_summary_json(datapoints):
     :return: JSON data
     :rtype: str
     """
-    summary_datapoints = [
-        _get_datapoint(datapoints, 'cpu.total'),
-        _get_datapoint(datapoints, 'memory.total'),
-        _get_datapoint(datapoints, 'memory.free'),
-        _get_datapoint(datapoints, 'filesystem.capacity.total', {'path': '/'}),
-        _get_datapoint(datapoints, 'filesystem.capacity.used', {'path': '/'})
+
+    metrics = [
+        ['cpu.total'],
+        ['memory.total'],
+        ['memory.free'],
+        ['filesystem.capacity.total', {'path': '/'}],
+        ['filesystem.capacity.used', {'path': '/'}],
     ]
+
+    summary_datapoints = []
+    for m in metrics:
+        datapoint = _get_datapoint(datapoints, *m)
+        if datapoint is not None:
+            summary_datapoints.append(datapoint)
+
     return json.dumps(summary_datapoints)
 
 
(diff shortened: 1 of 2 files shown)


klueska:
run all integration tests

klueska:

klueska:

python/lib/dcoscli/dcoscli/metrics.py
```diff
@@ -95,13 +95,30 @@ def _node_summary_json(datapoints):
...
+    summary_datapoints = []
+    for m in metrics:
+        datapoint = _get_datapoint(datapoints, *m)
+        if datapoint is None:
+            datapoint = {
+              "name": m[0],
+              "value": "N/A",
+              "unit": "N/A",
+              "timestamp": "N/A"
+            }
```
bamarni: When the datapoint is missing, is there any reason for adding an entry with `N/A` fields instead of omitting it from the list? This might break some scripts looping over the json and trying to parse units/value/timestamp as `N/A` is not a standard / well-defined way to indicate missing data.

bamarni:
IMO the summary in JSON should be an object where keys are the metric names and their values associated datapoints, as they are well known and a dict over a list should be easier to manipulate from the caller perspective.

As for the `N/A` values, I put them in tables as it's human readable output so we don't support scripting on it. For JSON I assume it'd break less likely if fields were set to null instead or the entry omitted.

In any case this PR is still better than a stack trace :slightly_smiling_face: So it's fine for me to have it as is too.

klueska:
What do you mean by the entry being "null"?

klueska:
Ah, I see. Yeah, I like that better. I will do that.

klueska:
@mesosphere-ci run windows integration tests

klueska:

ArmandGrillet:
@klueska Can you please rebase this PR? Also we have noticed a similar bug in `test_node_metrics_agent_details_json`:
```
tests/integrations/test_node.py::test_node_metrics_agent_details_json FAILED [ 65%]

_____________________ test_node_metrics_agent_details_json _____________________

    def test_node_metrics_agent_details_json():
        first_node_id = _node()[0]['id']
    
        node_json = fetch_valid_json(
            ['dcos', 'node', 'metrics', 'details', first_node_id, '--json']
        )
    
        names = [d['name'] for d in node_json]
>       assert 'system.uptime' in names
E       AssertionError: assert 'system.uptime' in []

tests/integrations/test_node.py:153: AssertionError
```

ArmandGrillet:

ArmandGrillet:
Let's use https://pypi.org/project/retrying/ in the tests.

klueska:

PR #42, opened 2018-09-25
Improved error message when shutting down an unexisting service.

ArmandGrillet:
https://jira.mesosphere.com/browse/DCOS_OSS-1273

### DIFF
--- a/python/lib/dcoscli/dcoscli/service/main.py
+++ b/python/lib/dcoscli/dcoscli/service/main.py
@@ -110,8 +110,13 @@ def _shutdown(service_id):
     :rtype: int
     """
 
-    mesos.DCOSClient().shutdown_framework(service_id)
-    return 0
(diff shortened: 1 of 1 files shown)


python/lib/dcoscli/dcoscli/service/main.py
```diff
@@ -110,9 +110,13 @@ def _shutdown(service_id):
     :rtype: int
     """
 
-    mesos.DCOSClient().shutdown_framework(service_id)
-    return 0
-
+    services = mesos.get_master().frameworks(inactive=False, completed=False)
+    for service in services:
+        if service.dict()["id"] == service_id:
```
bamarni: Are we confident that this key is always present in the dict? Otherwise it would raise an exception, unless it is retrieved with `service.dict().get('id')`

PR #57, opened 2018-10-09
Replace accidentally removed import

br-lewis:


scripts/publish_plugins.py
```diff
@@ -3,6 +3,7 @@
 import boto3
 
 from plugin import package_plugin
+import os
```
bamarni: this should be on the very top, as it's a stdlib import

bamarni: see https://www.python.org/dev/peps/pep-0008/ : 
```
Imports should be grouped in the following order:

Standard library imports.
Related third party imports.
Local application/library specific imports.
You should put a blank line between each group of imports.
```

PR #65, opened 2018-10-15
Static completions for the Core CLI

br-lewis:


bamarni:
@br-lewis : how could I check this locally? Is it compatible with the current Go CLI?

br-lewis:
Yeah. It actually requires the Go CLI. There are helper functions defined in `dcos completion bash` that these expect to exist.

armandgrillet:
Is there any way to have tests for this PR?

br-lewis:
I'm not sure how we would test this. I have the very basic testing script for completion in the wrapper CLI but so far we haven't filled that out much nor is it automatically run by anything, it's more a proof of concept at the moment.

These completion scripts are also dependent on code in the wrapper's completion script. That isn't entirely necessary (except for the entry points into the top level commands) but does help to reduce code duplication some. I think testing these completions would need to be part of an integration level test to like make sure plugins install properly but I'm really not sure.

armandgrillet:
@br-lewis OK for squash and merge?

PR #69, opened 2018-10-22
Added an 'index.html' file to represent the index of all CLI artifacts.

klueska:
This PR adds files to the `scripts` folder for publishing a list of all CLI artifacts to S3. The artifacts themselves are scraped from `downloads.dcos.io/cli/releases` and published as an `artifacts.json` file that is then read and rendered by a small html/css/javascript app.

A live version of the index can be seen here:
https://downloads.dcos.io/cli/index.html

Two different script are used to publish the necessary artifacts for this index.

The first is `publish_index.py` which publishes all of the necessary files required to render the index. This script should not need to be ran very often, and is only necessary if changes are made to any of the files used to implement the rendering of the index.

The second is `publish_artifacts.py` which publishes a JSON file listing all of the artifacts that should be displayed in the index. This script should be ran *every time* we publish a new artifact to S3 which we want to publish as a new release. This applies to both CLI releases themselves, as well as plugins.

In a subsequent commit we should add the calling of `publish_artifacts.py` to our CI as a step in publishing any of our release artifacts.


scripts/index/index.html
```diff
@@ -0,0 +1,44 @@
...
+
+<script type="text/javascript" src="html/jquery-3.3.1.min.js"></script>
+<script type="text/javascript" src="html/artifacts.js"></script>
+<script type="text/javascript" src="html/browser.js"></script>
+
+<title id="title"></title>
+</head>
+
+<body>
+<h1 id="header"></h1>
```
armandgrillet, line 18: We should uses classes instead of IDs for `header`, `names`, `versions`, and `listings` as they are used multiple times in this page.

scripts/index/html/artifacts.js
```diff
@@ -0,0 +1,354 @@
...
+
+      // List all binary artifacts.
+      this.list = function() {
+        return this.binaries;
+      };
+
+      // List unique major versions across all binaries.
+      this.majorVersions = function() {
+        var versions = [];
+        $.each(this.binaries, function(_, binary) {
```
armandgrillet, line 155: All occurrences of `$.each(a, function(b, c) {` should be replaceable by `a.forEach(function(c, b){`

scripts/index/publish_artifacts.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
+
+import os
+import boto3
```
armandgrillet, line 6: How are we getting `boto3`?

scripts/index/publish_artifacts.py
```diff
@@ -0,0 +1,86 @@
...
+import json
```
armandgrillet: The import should happen before `import boto3` and also `import os`. Adding a blank line between the stdlib/3rdparty imports would be great.

armandgrillet (approved):
LGTM. Using jQuery could have been avoided but it's making the code simpler, having multiple elements with the same ID in `index.html` is not best practice but it doesn't break the code. 
I would ship this PR if it is urgent, otherwise please check my comments.

scripts/index/index.html
klueska: They really need to be uniquely identifiable though. I could give them unique names if that's clearer. I just always reference them as e.g. `#binaries #header` in CSS or JS code though. Otherwise it would look like `#binaries #binaries-header`.

scripts/index/publish_artifacts.py
klueska: I'm not sure, but all of Bilal's other python scripts use this too without any sort of obvious virtualenv pulling it in. I assume the CI machines have it builtin.

scripts/index/index.html
klueska: What if I were to use "name" like in this answer:
https://stackoverflow.com/questions/9680037/how-do-i-select-an-element-with-its-name-attribute-in-jquery

Is that better?

I don't want to use "class" because I normally think of classes being used to apply style, which is not the case here.

scripts/index/html/artifacts.js
klueska: Is there a preference? I've often seen $.each() used all over the place in favor of forEach, but I'm not sure why. Can you explain and / or provide a link?

scripts/index/index.html
armandgrillet: Yes, `name` would be a better fit.

scripts/index/html/artifacts.js
armandgrillet: It is a preference, for this index jQuery could be dropped as you can see in http://youmightnotneedjquery.com/ (nearly all of the jQuery helper functions now exist in vanilla JS). As it would require refactoring and the JS code of this PR is clean, I will resolve this conversation and I don't think that changing any `$` calls is necessary.

klueska:
@bamarni do you think this script belongs in this repo or elsewhere? It needs to be run anytime there is a new artifact published for binaries and plugins (including the EE CLI).

bamarni:
I'd rather see it in dcos/dcos-cli as it's a more generic cli repo.

klueska:
Closed in favor of https://github.com/dcos/dcos-cli/pull/1376

PR #97, opened 2018-11-22
Normalize domains to lowercase when comparing URLs

bamarni:
Hostnames are case-insensitive and this caused issues with the `dcos
package install` command and CLI resources, as the auth token wasn't
being sent when the configured cluster URL and CLI resource URL had a
case mismatch on the domain.

https://jira.mesosphere.com/browse/DCOS_OSS-4501
https://jira.mesosphere.com/browse/DCOS-45391

### DIFF
--- a/dcos/http.py
+++ b/dcos/http.py
@@ -61,7 +61,7 @@ def _is_request_to_dcos(url, toml_config=None):
     # request should match scheme + netloc
     def _request_match(expected_url, actual_url):
         return expected_url.scheme == actual_url.scheme and \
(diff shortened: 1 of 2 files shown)


armandgrillet:
The AppVeyor issue, `The build phase is set to "MSBuild" mode (default), but no Visual Studio project or solution files were found in the root directory. If you are not building Visual Studio project switch build mode to "Script" and provide your custom build command.` has already been seen before.

armandgrillet (approved):

PR #108, opened 2018-12-06
Add delay subcommand for marathon subcommand

takirala:
JIRA issues: [DCOS_OSS-4325](https://jira.mesosphere.com/browse/DCOS_OSS-4325)

- Add new subcommand `dcos marathon delay reset` to reset the delay on a given appId or podId.

Sample responses:
```
$ dcos marathon delay reset sleep
Delay on [sleep] has been reset
$ dcos marathon delay reset /sleepasd
Error: Application /sleepasd not found in tasks queue.
```

NOTE : This should **NOT** be backported to 1.12 branch.

### DIFF
--- a/python/lib/dcos/dcos/marathon.py
+++ b/python/lib/dcos/dcos/marathon.py
@@ -446,6 +446,21 @@ class Client(object):
         response = self._rpc.http_req(http.post, path, params=params)
         return response.json()
 
+    def delay_reset(self, app_id):
+        """
+        :param app_id: The app_id which is used to delete the delay
+        :return: The result of reset delay call
+        :rtype: dict | str
+        """
+        app_id = util.normalize_marathon_id_path(app_id)
+        response = self._rpc.http_req(
+            http.delete,
+            'v2/queue{}/delay'.format(app_id)
+        )
+        if response.status_code == 204:
+            return 'Delay on [{}] has been reset'.format(app_id)
+        raise DCOSHTTPException(response)
+
     def get_deployment(self, deployment_id):
         """Returns a deployment.
 
--- a/python/lib/dcos/dcos/marathon.py
+++ b/python/lib/dcos/dcos/marathon.py
@@ -446,19 +446,19 @@ class Client(object):
         response = self._rpc.http_req(http.post, path, params=params)
         return response.json()
 
-    def delay_reset(self, app_id):
+    def delay_reset(self, id):
         """
-        :param app_id: The app_id which is used to delete the delay
+        :param id: The app_id or pod_id which is used to delete the delay
         :return: The result of reset delay call
         :rtype: str
         """
-        app_id = util.normalize_marathon_id_path(app_id)
+        id = util.normalize_marathon_id_path(id)
         response = self._rpc.http_req(
             http.delete,
-            'v2/queue{}/delay'.format(app_id)
+            'v2/queue{}/delay'.format(id)
         )
         if response.status_code == 204:
-            return 'Delay on [{}] has been reset'.format(app_id)
+            return 'Delay on [{}] has been reset'.format(id)
         raise DCOSHTTPException(response)
 
     def get_deployment(self, deployment_id):
--- a/python/lib/dcos/dcos/marathon.py
+++ b/python/lib/dcos/dcos/marathon.py
@@ -450,7 +450,7 @@ class Client(object):
         """
         :param app_id: The app_id which is used to delete the delay
         :return: The result of reset delay call
-        :rtype: dict | str
+        :rtype: str
         """
         app_id = util.normalize_marathon_id_path(app_id)
         response = self._rpc.http_req(
--- a/python/lib/dcoscli/dcoscli/data/help/marathon.txt
+++ b/python/lib/dcoscli/dcoscli/data/help/marathon.txt
@@ -17,6 +17,7 @@ Usage:
     dcos marathon app kill [--scale] [--host=<host>] <app-id>
     dcos marathon app update [--force] <app-id> [<properties>...]
     dcos marathon app version list [--max-count=<max-count>] <app-id>
+    dcos marathon delay reset <app-id>|<pod-id>
     dcos marathon deployment list [--json|--quiet] [<app-id>]
     dcos marathon deployment rollback <deployment-id>
     dcos marathon deployment stop <deployment-id>
@@ -69,6 +70,8 @@ Commands:
         Update an application.
     app version list
         List the version history of an application.
+    delay reset
+        Reset the current delay (if any) of the application.
     deployment list
         Print a list of currently deployed applications.
     deployment rollback
--- a/python/lib/dcoscli/dcoscli/marathon/main.py
+++ b/python/lib/dcoscli/dcoscli/marathon/main.py
@@ -230,7 +230,7 @@ def _cmds():
 
         cmds.Command(
             hierarchy=['marathon', 'delay', 'reset'],
-            arg_keys=['<app-id>', '<pod-id>'],
+            arg_keys=['<app-id>'],
             function=subcommand.delay_reset),
 
         cmds.Command(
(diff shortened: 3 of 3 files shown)


takirala:

takirala:

takirala:

takirala:
@mesosphere-ci run all integration tests please

takirala:

armandgrillet:
This looks good to me, is it planned to add tests? Asking as I see the related ticket as `In Review`.

takirala:
- The new functionality does not do any validation other than printing whatever the upstream responds.
- It is difficult to add integration tests until we make a release for marathon endpoint.
- There are good amount of tests added to upstream via https://github.com/mesosphere/marathon/pull/6700/files#diff-c0ddfcb6ba2bc74c0d863fee0cf62951R88 (I added integration tests especially keeping in mind CLI codebase would lack integration tests until the feature is released to `dcos-enterprise`).
- Adding unit tests did not seem valuable as the subcommand isn't doing much other than printing whatever content upstream returns without any modifications or such.

cc @ArmandGrillet @kensipe wdyt ?

armandgrillet (approved):

python/lib/dcos/dcos/marathon.py
```diff
@@ -446,6 +446,31 @@ def restart_app(self, app_id, force=False):
...
+
+    def delay_show(self, run_spec_id):
+        """
+        :param run_spec_id: Shows the delay for specified run_spec_id
+        :return: The result of get delay call
+        :rtype: dict
+        """
+        return self._rpc.http_req(
+            http.get,
+            'v2/queue/{}/delay'.format(run_spec_id)
```
kensipe: I would rather stick to the standard in this file in creating a `path` var.

example: https://github.com/dcos/dcos-core-cli/blob/d5e0f27f57700e5bdcdfe70c84d1a7154e7a617a/python/lib/dcos/dcos/marathon.py#L444

python/lib/dcos/dcos/marathon.py
```diff
@@ -446,6 +446,31 @@ def restart_app(self, app_id, force=False):
...
+        :return: The result of reset delay call
+        :rtype: dict | str
+        """
+        response = self._rpc.http_req(
+            http.delete,
+            'v2/queue/{}/delay'.format(run_spec_id)
+        )
+        if response.status_code == 204:
+            return 'Delay on [{}] has been reset'.format(run_spec_id)
+        return response.text
```
kensipe: what are the possibilities here?    Are you sure it is JSON `dict`?   
we have other (although) few examples of returning different `rtype`s.   however they are `.json()` and are a collection of ids...  not just any random message.

digging into this...    the response is `NoContent` which is 204 https://github.com/mesosphere/marathon/blob/master/src/main/scala/mesosphere/marathon/api/v2/QueueResource.scala#L52

So the response is either Authorized with a 204 and no text.   or it is not authorized...

python/lib/dcoscli/dcoscli/data/help/marathon.txt
```diff
@@ -69,6 +71,10 @@ Commands:
         Update an application.
     app version list
         List the version history of an application.
+    delay show
+        Show the current delay (if any) of the application.
```
kensipe: I don't see value in `(if any)`

python/lib/dcoscli/dcoscli/data/help/marathon.txt
```diff
@@ -17,6 +17,8 @@ Usage:
     dcos marathon app kill [--scale] [--host=<host>] <app-id>
     dcos marathon app update [--force] <app-id> [<properties>...]
     dcos marathon app version list [--max-count=<max-count>] <app-id>
+    dcos marathon delay reset <app-id>|<pod-id>
+    dcos marathon delay show <app-id>|<pod-id>
```
kensipe: I believe the `|` in `<app-id>|<pod-id>` will restrict to 1 argument... and the underlying code will only get 1.   I think we need to be generic here and provide details in the help section of this file.

perhaps a <app-pod-id> with details below

python/lib/dcoscli/dcoscli/marathon/main.py
```diff
@@ -228,6 +228,16 @@ def _cmds():
...
             function=subcommand.debug_details),
 
+        cmds.Command(
+            hierarchy=['marathon', 'delay', 'show'],
+            arg_keys=['<app-id>', '<pod-id>'],
+            function=subcommand.delay_show),
+
+        cmds.Command(
+            hierarchy=['marathon', 'delay', 'reset'],
+            arg_keys=['<app-id>', '<pod-id>'],
```
kensipe: I really dislike the 2 ids as mentioned above

python/lib/dcoscli/dcoscli/marathon/main.py
```diff
@@ -856,6 +867,34 @@ def version_list(self, app_id, max_count):
         emitter.publish(versions)
         return 0
 
+    def delay_reset(self, app_id, pod_id):
```
kensipe: only 1 argument is ever possible... this will just confuse future developers IMO

python/lib/dcoscli/dcoscli/marathon/main.py
```diff
@@ -856,6 +867,34 @@ def version_list(self, app_id, max_count):
...
+        :param app_id: the id of the application to reset delay
+        :type app_id: str
+        :param pod_id: the Marathon ID of the pod to reset delay
+        :type pod_id: str
+        :return: process return code
+        :rtype: int
+        """
+        response = self._create_marathon_client().delay_reset(app_id or pod_id)
+        if response:
+            emitter.publish(response)
```
kensipe: this is either a string or a dict... while we do emit JSON... it is only when the user uses the `--json` flag.

Also... it should NOT be a `return 0` below if this is a dict... only if a string.

kensipe: although there are a number of places where this code naively returns 0... I believe that is because an exception could be thrown... in which case python will caught the exception and return a non-zero.   (I would have to dig deeper to confirm).   This would not be the case in this code.   I believe the code flow might be easier if you have `204` return a value and anything else throw an exception.  Those the return of that function is only a `str`.

kensipe (changes requested):

python/lib/dcoscli/dcoscli/data/help/marathon.txt
takirala: @kensipe
What we are passing here is a `runSpecId`. But there is no `runSpecId` (yet) in the help menu. Creating `runSpecId` would address all other concerns you mentioned in trailing comments, but I am not sure if we want to do that. What else do you suggest here?

python/lib/dcos/dcos/marathon.py
```diff
@@ -21,6 +21,7 @@ def create_client(toml_config=None):
         toml_config = config.get_config()
 
     marathon_url = _get_marathon_url(toml_config)
+    marathon_url = "http://localhost:8080"
```
kensipe: this should removed

kensipe (changes requested):

python/lib/dcos/dcos/marathon.py
```diff
@@ -446,6 +446,21 @@ def restart_app(self, app_id, force=False):
         response = self._rpc.http_req(http.post, path, params=params)
         return response.json()
 
+    def delay_reset(self, app_id):
+        """
+        :param app_id: The app_id which is used to delete the delay
```
kensipe: this doc should be updated to refer to pod_id or app_id.  Also... the `app_id` is required for `marathon/main.py` but this id arg name could be changed to reflect what we want.   it might be best to call it `id`

python/lib/dcoscli/dcoscli/marathon/main.py
```diff
@@ -856,6 +862,24 @@ def version_list(self, app_id, max_count):
...
         return 0
 
+    def delay_reset(self, app_id):
+        """
+        :param app_id: the id of the application to reset delay.
+                       This can be a pod_id as well.
+        :type app_id: str
+        :rtype: int
+
+        NOTE: This method takes ONLY one parameter `app_id` but the help menu says that
```
kensipe: ❤️

kensipe (approved):
nit in marathon client.. but not a show stopper... lgtm

takirala:

takirala:
Thanks for the review @kensipe

armandgrillet:

armandgrillet:
@mesosphere-ci run windows integration tests

PR #112, opened 2018-12-07
[1.12] Fixed date string parsing.

rgo3:
With this commit we fix the date string parsing in the `dcos-core-cli`
plugin to parse more combinations of the ISO 8601 standard. This issue
was seen in when the datestring had a different format than
`2017-03-30T15:50:16.187+0000`.

https://jira.mesosphere.com/browse/DCOS-45992

### DIFF
--- a/python/lib/dcoscli/dcoscli/tables.py
+++ b/python/lib/dcoscli/dcoscli/tables.py
@@ -7,6 +7,7 @@ import textwrap
 
 from collections import OrderedDict
 
+import dateutil.parser
 import prettytable
 
 from dcos import auth, marathon, mesos, util
@@ -422,8 +423,8 @@ def _str_to_datetime(datetime_str):
     """
     if not datetime_str:
         return None
-    datetime_str = datetime_str.split('+')[0]
-    return datetime.datetime.strptime(datetime_str, "%Y-%m-%dT%H:%M:%S.%f")
+    # Used to parse ISO 8601 formatted date strings.
+    return dateutil.parser.parse(datetime_str)
 
 
 def _last_run_status(job):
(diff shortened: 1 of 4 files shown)


rgo3:
@mesosphere-ci  run all integration tests

rgo3:

rgo3:

python/lib/dcoscli/dcoscli/tables.py
```diff
@@ -422,8 +423,8 @@ def _str_to_datetime(datetime_str):
     """
     if not datetime_str:
         return None
-    datetime_str = datetime_str.split('+')[0]
-    return datetime.datetime.strptime(datetime_str, "%Y-%m-%dT%H:%M:%S.%f")
+    # Used to parse ISO 8601 formatted date strings.
+    return dateutil.parser.parse(datetime_str)
```
bamarni, line 427: Can it be done without an extra package? eg. https://docs.python.org/3.5/library/datetime.html#datetime.date.isoformat

python/lib/dcoscli/tests/unit/test_tables.py
```diff
@@ -192,3 +193,21 @@ def _test_table(table_fn, fixture_fn, path):
     table = table_fn(fixture_fn)
     with open(path) as f:
         assert str(table) == f.read().strip('\n')
+
+
+def test_str_to_datetime():
```
bamarni, line 198: :+1:

python/lib/dcoscli/dcoscli/tables.py
rgo3: So what we need is a function that parses an ISO 8061 string and returns a `datetime` object. I found this https://docs.python.org/3.5/library/datetime.html#datetime.datetime.isoformat but it does exactly the opposite of what we need, turning a `datetime` object into a string. I have not found anything in pythons standard library that would be able to parse the variety of ISO 8061 formatted strings like `dateutil.parser.parse()`.

bamarni (approved):

PR #113, opened 2018-12-10
[1.11] Fixed date string parsing.

rgo3:
With this commit we fix the date string parsing in the `dcos-core-cli`
plugin to parse more combinations of the ISO 8601 standard. This issue
was seen in when the datestring had a different format than
`2017-03-30T15:50:16.187+0000`.

https://jira.mesosphere.com/browse/DCOS-45992


armandgrillet:

PR #125, opened 2019-01-08
[1.13] Update metric names

gracedo:
https://jira.mesosphere.com/browse/DCOS_OSS-4622
The metrics upgrade from 1.11 to 1.12 resulted in updated metric names in the metrics API, which were not reflected here in the CLI which hits the metrics API for task metrics. I've tested this fix locally and it is now populating the CPU field (MEM and DISK metrics were not available in both versions to begin with so they are 0s here).

Previously:
```
CPU           MEM              DISK             
0.00 (0.00%)  0.00GiB (0.00%)  0.00GiB (0.00%)
```
Now:
```
CPU                MEM              DISK             
20534.96 (99.21%)  0.00GiB (0.00%)  0.00GiB (0.00%)
```

### DIFF
--- a/python/lib/dcoscli/dcoscli/metrics.py
+++ b/python/lib/dcoscli/dcoscli/metrics.py
@@ -156,9 +156,9 @@ def _task_summary_json(datapoints):
     :rtype: str
     """
     summary_datapoints = [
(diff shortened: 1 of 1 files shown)


gracedo:

armandgrillet (approved):

PR #126, opened 2019-01-08
[1.12] Update metric names

gracedo:
[backport to 1.12 of #125]

https://jira.mesosphere.com/browse/DCOS_OSS-4622
The metrics upgrade from 1.11 to 1.12 resulted in updated metric names in the metrics API, which were not reflected here in the CLI which hits the metrics API for task metrics. I've tested this fix locally and it is now populating the CPU field (MEM and DISK metrics were not available in both versions to begin with so they are 0s here).

Previously:
```
CPU           MEM              DISK             
0.00 (0.00%)  0.00GiB (0.00%)  0.00GiB (0.00%)
```
Now:
```
CPU                MEM              DISK             
20534.96 (99.21%)  0.00GiB (0.00%)  0.00GiB (0.00%)
```

### DIFF
--- a/python/lib/dcoscli/dcoscli/metrics.py
+++ b/python/lib/dcoscli/dcoscli/metrics.py
@@ -156,9 +156,9 @@ def _task_summary_json(datapoints):
     :rtype: str
     """
     summary_datapoints = [
-        _get_datapoint(datapoints, 'cpus.user.time'),
-        _get_datapoint(datapoints, 'mem.total'),
-        _get_datapoint(datapoints, 'disk.used')
+        _get_datapoint(datapoints, 'cpus.user_time_secs'),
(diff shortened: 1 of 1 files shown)


armandgrillet (approved):

armandgrillet:

gracedo:
🙌 thanks @ArmandGrillet ! Had a couple of questions re: this work - when will the team cut the next release of the CLI so I know when this fix will go out? Do you have any suggestions of where/how to add integration tests (e.g. in DCOS?) that will catch this in the future if we update metric names again?

armandgrillet:
I suggest you adding an integration test to `python/lib/dcoscli/tests/integrations/test_node.py`. We can cut a release whenever we want, will update the related COPS ticket when this happens.

PR #130, opened 2019-01-10
Print `run-id` when starting a new job run.

rgo3:


rgo3:

rgo3:

pkg/cmd/job/job_run.go
```diff
@@ -17,7 +19,8 @@ func newCmdJobRun(ctx api.Context) *cobra.Command {
 				return err
 			}
 
-			_, err = client.RunJob(args[0])
+			run, err := client.RunJob(args[0])
+			fmt.Fprintln(ctx.Out(), run.ID)
```
bamarni, line 26: we should check for error here and return early, as "run" might be nil.

rgo3:

PR #154, opened 2019-01-23
Added Go subcommand `dcos node list`.

rgo3:
With this commit we add `dcos node list` to the `dcos node` top-level
command. It deprecates the functionality of `dcos node` to show a list
of nodes. From now on, if `DCOS_CLI_STRICT_DEPRECATIONS` is set,
`dcos node` will only display the help menu for that command.


pkg/mesos/client.go
```diff
@@ -0,0 +1,90 @@
...
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	switch resp.StatusCode {
+	case 200:
+		var state State
+		err = json.NewDecoder(resp.Body).Decode(&state)
+		return &state, err
+	case 503:
```
mrnugget, line 52: Is `503` really a special case? I think checking for just `200` should be enough (which is also what you're doing in `Masters()`.

mrnugget (commented):
Not too familiar with the conventions you're using yet, so I can't comment on layout/organization.

But: no tests? :)

pkg/mesos/client.go
rgo3: I have that status code from the Mesos [docs](http://mesos.apache.org/documentation/latest/endpoints/master/state/) and added it as a special case to have a more specific error message if this happens.

pkg/cmd/node/node.go
```diff
@@ -0,0 +1,62 @@
...
+
+	baseURL, _ := cluster.Config().Get("core.dcos_url").(string)
+	if baseURL == "" {
+		baseURL = cluster.URL()
+	}
+
+	return mesos.NewClient(
+		httpclient.New(
+			baseURL,
+			httpclient.Logger(ctx.Logger()),
```
bamarni, line 51: Here the logger will never display HTTP logs, one thing we should do is have a helper function to create an httpclient for plugins, that will support the `DCOS_VERBOSITY` env var which gets passed to plugins on invocation.

pkg/mesos/client.go
```diff
@@ -0,0 +1,90 @@
...
+	"fmt"
+
+	"github.com/dcos/dcos-cli/pkg/httpclient"
+	"github.com/sirupsen/logrus"
+)
+
+// Client is a Mesos client for DC/OS.
+type Client struct {
+	http   *httpclient.Client
+	logger *logrus.Logger
```
bamarni: This field doesn't seem to be used.

pkg/mesos/types.go
```diff
@@ -0,0 +1,97 @@
...
+	Type                string                 `json:"type"`
+	Version             string                 `json:"version"`
+	Zone                string                 `json:"zone"`
+}
+
+// IP returns the IP of a slave parsed from its PID.
+func (s *Slave) IP() string {
+	// PID format: Host@IP:Port
+	s1 := strings.Split(s.PID, "@")
+	s2 := strings.Split(s1[1], ":")
```
bamarni, line 74: Is there a simpler way to get slaves IPs?

rgo3: Keeping it for now, to avoid backwards compatibility issues.

PR #156, opened 2019-01-24
[CI] Pass the --update flag when adding the core plugin

bamarni:
Otherwise this causes the command to fail with a "dcos-core-cli is
already installed" error, as it gets installed during cluster setup now.


bamarni:

bamarni:

PR #159, opened 2019-01-28
Added Go subcommand `dcos node list-components`.

armandgrillet:


pkg/cmd/node/node_list.go
```diff
@@ -20,7 +20,7 @@ func newCmdNodeList(ctx api.Context) *cobra.Command {
 	var jsonOutput bool
 	cmd := &cobra.Command{
 		Use:   "list",
-		Short: "Show all nodes in the cluster",
+		Short: "Show all nodes in the cluster.",
```
rgo3: In help menus we don't have a `.` at the end of the description
```suggestion
		Short: "Show all nodes in the cluster",
```

pkg/cmd/node/node_list_components.go
```diff
@@ -0,0 +1,98 @@
...
+						ip = agent.AgentInfo.GetHostname()
+					}
+				}
+				if ip == "" {
+					return fmt.Errorf("agent '%s' not found", mesosID)
+				}
+			}
+
+			// Reuse the existing HTTP client of the Mesos client instead of creating a new one.
+			resp, err := client.HTTP().Get(fmt.Sprintf("/system/health/v1/nodes/%s/units", ip))
```
rgo3: Is this the only place in the CLI a call to this API is made? I also don't think it belongs in the `mesos` package, but having it in the `RunE()` func of the command while still using the mesos client seems not clean to me. This and the respective type should go in a separate package (`health` maybe?) IMO. Will make the `RunE()` more readable and shorter.

pkg/mesos/client.go
```diff
@@ -22,7 +22,29 @@ func NewClient(baseClient *httpclient.Client) *Client {
 	}
 }
 
-// Masters is returning the mesos masters of the connected cluster.
+// HTTP returns the HTTP client of the Mesos client.
+func (c *Client) HTTP() *httpclient.Client {
```
rgo3: I think we don't need this if we do a correct separation of concerns in the calling function.

pkg/mesos/types.go
```diff
@@ -89,6 +90,14 @@ type State struct {
 	Slaves            []Slave `json:"slaves"`
 }
 
+// GetAgents represents the JSON returned when using GET_AGENTS.
+type GetAgents struct {
```
rgo3: I think we can come up with a better naming scheme here.

pkg/cmd/node/node_list_components.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/spf13/cobra"
+)
+
+func newCmdNodeListComponents(ctx api.Context) *cobra.Command {
+	var jsonOutput bool
+	var leader bool
+	var mesosID string
+	cmd := &cobra.Command{
+		Use:   "list-components",
+		Short: "Print a list of available DC/OS components on specified node.",
```
rgo3:
```suggestion
		Short: "Print a list of available DC/OS components on specified node",
```

pkg/cmd/node/node_list_components.go
```diff
@@ -0,0 +1,98 @@
...
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if !leader && mesosID == "" {
```
rgo3, line 22: The checking of correct flag usage can go into a `PreRunE()`.

pkg/mesos/client.go
```diff
@@ -22,7 +22,29 @@ func NewClient(baseClient *httpclient.Client) *Client {
...
+	return c.http
+}
+
+// Leaders returns the mesos leaders of the connected cluster.
+func (c *Client) Leaders() ([]Master, error) {
```
rgo3: Should this function maybe only return one Master? I know it is possible to get multiple but is that ever useful? Should Mesos ever have multiple leaders? Otherwise I would suggest to already check here if multiple leaders are returned and if so throw an error. Unless it is useful to have multiple leaders, this would save checking for that case when this function is called.

rgo3 (changes requested):

pkg/cmd/node/node_list.go
armandgrillet: True, got confused because we did that in the old help menus.

pkg/mesos/types.go
```diff
@@ -1,8 +1,17 @@
 package mesos
 
 import "strings"
+import "github.com/mesos/mesos-go/api/v1/lib/master"
 
-// Master represents a single mesos master node
+// Agents represents the JSON returned when using GET_AGENTS.
+type Agents struct {
```
bamarni: We could directly use this type : https://github.com/mesos/mesos-go/blob/f27edc01ca67f9b7e1b6a4ea9a1086e5372eaf0c/api/v1/lib/master/master.pb.go#L932

pkg/mesos/client.go
```diff
@@ -79,14 +99,32 @@ func (c *Client) StateSummary() (*StateSummary, error) {
 	}
 }
 
-// Slaves returns the agents of the mesos cluster.
-func (c *Client) Slaves() ([]Slave, error) {
-	state, err := c.StateSummary()
+// Agents returns the agents of the mesos cluster.
+func (c *Client) Agents() ([]master.Response_GetAgents_Agent, error) {
```
bamarni, line 103: Currently we are only returning types defined in this package (`Master`, `State`, `StateSummary`, etc), thus the mesos-go dependency is not exposed and remains an implementation detailed. Can we follow the same approach here for consistency?

pkg/health/client.go
```diff
@@ -0,0 +1,40 @@
+package health
```
bamarni: As discussed, this should be a diagnostics package because the endpoint is exposed from the DC/OS diagnostics component.

pkg/health/types.go
```diff
@@ -0,0 +1,11 @@
+package health
+
+// Units represents the units of a node returned by the '/units' endpoint.
+type Units struct {
```
bamarni: To avoid duplicating the response definition, we should directly use https://github.com/dcos/dcos-diagnostics/blob/master/api/structures.go#L16

pkg/cmd/node/node_list_components.go
```diff
@@ -0,0 +1,75 @@
...
+					return fmt.Errorf("invalid leader response, missing field 'ip'")
+				}
+				ip = leader.IP
+			} else {
+				agents, err := mesosClient().Agents()
+				if err != nil {
+					return err
+				}
+				for _, agent := range agents {
+					if mesosID == agent.AgentInfo.GetHostname() {
```
bamarni: As discussed, comparing against the IP/hostname is a bug, the `--mesos-id` option description specifies it refers to the node ID.

pkg/mesos/client.go
bamarni: (Maybe we should do it the other way around, and update other methods to return "pure" definitions from the structs in `mesos-go`, we just need to decide on which approach to choose and stick to it.)

bamarni (approved):

pkg/mesos/client.go
bamarni: @rgo3 : FYI we're going to look at this separately from this PR.

armandgrillet:
We can see in CI:
```
../../../../pkg/mod/github.com/coreos/go-systemd@v0.0.0-20181031085051-9002847aa142/sdjournal/journal.go:27:33: fatal error: systemd/sd-journal.h: No such file or directory
 // #include <systemd/sd-journal.h>
```
This seems related to https://github.com/coreos/go-systemd/issues/116 and https://github.com/coreos/go-systemd/issues/201

rgo3 (approved):

PR #160, opened 2019-01-30
Added Go subcommand `dcos node metrics`.

armandgrillet:


pkg/cmd/node/node_metrics.go
```diff
@@ -0,0 +1,112 @@
...
+func newCmdNodeMetrics(ctx api.Context) *cobra.Command {
+	var jsonOutput bool
+	cmd := &cobra.Command{
+		Use:       "metrics",
+		Short:     "Print details or summary of the metrics of a specified agent",
+		ValidArgs: []string{"details", "summary"},
+		Args:      cobra.ExactArgs(2),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if args[0] != "details" && args[0] != "summary" {
+				return fmt.Errorf("the first argument must be 'details' or 'metrics' e.g. dcos node metrics details <mesos-id> [--json]")
```
bamarni: `'details' or 'metrics'` >  `'details' or 'summary'`

However, shall we rather create 2 subcommands (`details` and `summary`)? It removes the need for manual validation and separates the code (looking at the body of the command it looks like it's 2 totally different code paths).

pkg/metrics/client.go
```diff
@@ -0,0 +1,40 @@
...
+
+// NewClient creates a new metrics client.
+func NewClient(baseClient *httpclient.Client) *Client {
+	return &Client{
+		http: baseClient,
+	}
+}
+
+// Node returns the units of a certain node.
+func (c *Client) Node(mesosID string) (Node, error) {
```
rgo3:
```suggestion
func (c *Client) Node(mesosID string) (*Node, error) {
```

Would return a pointer here. Makes it more consistent throughout other API calls and instead of `Node{}` you can return `nil`.

rgo3 (approved):
Otherwise LGTM.

PR #165, opened 2019-02-04
Added public IP field when listing nodes.

armandgrillet:


pkg/networking/client.go
```diff
@@ -0,0 +1,54 @@
+package networking
```
bamarni, line 1: I think it should belong to the dcos client (https://github.com/dcos/dcos-cli/blob/master/pkg/dcos/client.go), though for now we don't have an easy way to share it between the 2 projects.

pkg/networking/client.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/dcos/dcos-cli/pkg/httpclient"
+)
+
+// Client is a diagnostics client for DC/OS.
```
bamarni: `networking`?

pkg/networking/client.go
```diff
@@ -0,0 +1,54 @@
...
+
+	ips := make(map[string][]string)
+	for _, node := range *nodes {
+		ips[node.PrivateIP] = node.PublicIps
+	}
+	return ips, nil
+}
+
+// nodes returns the nodes in a cluster.
+func (c *Client) nodes() (*[]Node, error) {
```
bamarni: Can this method be exported? As it exposes the information in the endpoint.

pkg/networking/client.go
```diff
@@ -0,0 +1,54 @@
...
+
+// NewClient creates a new networking client.
+func NewClient(baseClient *httpclient.Client) *Client {
+	return &Client{
+		http: baseClient,
+	}
+}
+
+// IPs returns the private and related public IPs of the nodes in a cluster.
+func (c *Client) IPs() (map[string][]string, error) {
```
bamarni: I'd put this function outside of the client, as it's custom logic used in `dcos node list` to match the table entries with the endpoint information.

pkg/networking/client.go
bamarni: Looking at https://docs.mesosphere.com/1.12/api/master-routes/ we have many different api sections in DC/OS, we need to decide which ones would have their own client, and which ones should be under a generic dc/os client.

pkg/cmd/node/node_list.go
```diff
@@ -26,50 +34,75 @@ func newCmdNodeList(ctx api.Context) *cobra.Command {
 			client := mesosClient()
 
 			// The following code allows us to do the API calls for
-			// the state and the masters concurrently which results
-			// in a significant speedup of the command.
-			res := make(chan stateResult)
+			// the state, the ips, and the masters concurrently which
```
rgo3:
```suggestion
			// the state, the public IPs, and the masters concurrently which
```

pkg/cmd/node/node_list.go
```diff
@@ -26,50 +34,75 @@ func newCmdNodeList(ctx api.Context) *cobra.Command {
 			client := mesosClient()
 
 			// The following code allows us to do the API calls for
```
rgo3: Should we use this PR to clean up the `RunE()´? Or do we revisit it when reviewing the node command? I think it is too long.

pkg/networking/client.go
```diff
@@ -0,0 +1,40 @@
...
+
+// NewClient creates a new networking client.
+func NewClient(baseClient *httpclient.Client) *Client {
+	return &Client{
+		http: baseClient,
+	}
+}
+
+// Nodes returns the nodes in a cluster.
+func (c *Client) Nodes() (*[]Node, error) {
```
rgo3: There is no need to return a pointer to the slice. See https://github.com/dcos/dcos-core-cli/blob/08bbf476ca0cd7a7126cce5092bec22a86ba4333/pkg/metronome/client.go#L100

pkg/cmd/node/node_list.go
armandgrillet: Let's do that in the next PR.

armandgrillet:
@mesosphere-ci run integration tests

rgo3 (approved):

PR #172, opened 2019-02-07
Refactored 'dcos node diagnostics' in Go.

armandgrillet:


pkg/diagnostics/client.go
```diff
@@ -38,3 +45,126 @@ func (c *Client) Units(node string) (*UnitsHealthResponseJSONStruct, error) {
...
+func (c *Client) GetBundle(bundle string) (*http.Response, error) {
+	if !strings.HasSuffix(bundle, ".zip") {
+		return nil, fmt.Errorf("Format allowed bundle-*.zip")
+	}
+
+	return c.http.Get("/system/health/v1/report/diagnostics/serve/" + bundle)
+}
+
+// DeleteBundle deletes a diagnostics bundle in the cluster.
+func (c *Client) DeleteBundle(bundle string) (*BundleGenericResponseJSONStruct, error) {
```
bamarni: `Delete()`? As diagnostics bundles are the primary entity for this API it seems implicit. For example there is another method called `List()` and not `ListBundles()`.

pkg/diagnostics/client.go
```diff
@@ -19,13 +23,16 @@ func NewClient(baseClient *httpclient.Client) *Client {
 	}
 }
 
+// DIAGNOSTICS_BASE_URL = '/system/health/v1/report/diagnostics/'
```
bamarni: leftover

pkg/cmd/node/node_diagnostics_status.go
```diff
@@ -0,0 +1,43 @@
...
+				enc := json.NewEncoder(ctx.Out())
+				enc.SetIndent("", "    ")
+				return enc.Encode(status)
+			}
+			for k, v := range status {
+				fmt.Println(k)
+				val := reflect.ValueOf(v)
+				t := val.Type()
+				for i := 0; i < t.NumField(); i++ {
+					fmt.Printf("  %s: ", t.Field(i).Tag.Get("json"))
```
bamarni, line 34: For the record, this would print unwanted data in case there are options like `omitempty` on the json tag. I think it'd be more idiomatic Go to add a `String()` method to the `BundleReportStatus`.

pkg/diagnostics/client.go
bamarni: `Create()` and `Cancel()`? for the same reasons

pkg/cmd/node/node_diagnostics.go
```diff
@@ -0,0 +1,55 @@
...
+				subCommand.SilenceUsage = true
+				subCommand.SetArgs(ctx.Args()[4:])
+				return subCommand.Execute()
+			}
+			return cmd.Help()
+		},
+	}
+	cmd.Flags().BoolVar(&cancel, "cancel", false, "Cancel a running diagnostics job")
+	cmd.Flags().MarkDeprecated("cancel", "use the 'cancel' subcommand instead")
+	cmd.Flags().BoolVar(&jsonOutput, "json", false, "Print in json format")
```
bamarni: this can be `cmd.Flags().Bool("json", false, "Print in json format")`, the variable is not needed as it's not being used, it's just for cobra to register the option.

pkg/cmd/node/node_diagnostics.go
```diff
@@ -0,0 +1,55 @@
...
+		Use:   "diagnostics",
+		Short: "Use diagnostics bundles",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var subCommand *cobra.Command
+			switch {
+			case cancel && !list && !status:
+				subCommand = newCmdNodeDiagnosticsCancel(ctx)
+			case !cancel && list && !status:
+				subCommand = newCmdNodeDiagnosticsList(ctx)
+			case !cancel && !list && status:
```
bamarni: It is harder to read when checking for mutual exclusion, usually when we do this we also print an informative error message. Given that it is also deprecated we should keep it simple, here I'd just check for each option individually and match the first one : 
``` go
if list {
    subCommand = newCmdNodeDiagnosticsList(ctx)
} else if status {
    // ...
} // ...
```

pkg/cmd/node/node_diagnostics_cancel.go
```diff
@@ -0,0 +1,34 @@
...
+			if err != nil {
+				return err
+			}
+
+			if jsonOutput {
+				enc := json.NewEncoder(ctx.Out())
+				enc.SetIndent("", "    ")
+				return enc.Encode(resp)
+			}
+			fmt.Println(resp.Status)
```
bamarni, line 28: With the python CLI, is it printing the HTTP response status code in case of success? Do we still want to do this?

pkg/cmd/node/node_diagnostics_status.go
rgo3: Regarding whats being printed here overall, yesterday I talked to @ArmandGrillet and mentioned it might be worth redesigning the output. What is the benefit of having this mixed json/text output and not a table like everywhere else?

pkg/cmd/node/node_diagnostics_cancel.go
armandgrillet: Yes it does with the Python CLI:
```
➜  dcos-core-cli (DCOS_OSS-4697) ✔ dcos node diagnostics --cancel
Job is not running
```
`Job is not running` is the status of the response. I think it is useful as if a job is not running we have not canceled anything.

pkg/cmd/node/node_diagnostics_status.go
bamarni: I think a table would be unreadable because of some fields being quite long and was probably the reason for choosing this approach. We will have a ticket about reviewing the new node subcommand and its UX on the next sprints, where we can consider changing the output.

bamarni (approved):
I think it's fine from the refactoring perspective. We will still have to double check the UX and change a few things as a second step.

PR #174, opened 2019-02-11
Added Go subcommand 'dcos node ssh'.

rgo3:
This includes a sshclient package that configures a SSH client to be
able to establish connections to all the nodes of a DC/OS cluster.

### DIFF
--- a/pkg/cmd/node/node_ssh.go
+++ b/pkg/cmd/node/node_ssh.go
@@ -1,0 +1,136 @@
+package node
+
+import (
+	"fmt"
+
+	"github.com/dcos/dcos-cli/pkg/dcos"
+	"github.com/dcos/dcos-core-cli/pkg/pluginutil"
+	"github.com/dcos/dcos-core-cli/pkg/sshclient"
+
+	"github.com/dcos/dcos-cli/api"
+	"github.com/spf13/cobra"
+)
+
+func newCmdNodeSSH(ctx api.Context) *cobra.Command {
+	var leader, masterProxy bool
+	var mesosID, proxyIP, privateIP string
+
+	clientOpts:= sshclient.ClientOpts{
+		Input:      ctx.Input(),
+		Out:        ctx.Out(),
+		ErrOut:     ctx.ErrOut(),
+	}
+
+	cmd := &cobra.Command{
+		Use:   "ssh <command>",
+		Short: "Establish an SSH connection to the master or agent nodes of your DC/OS cluster",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var err error
+			clientOpts.Host, err = detectHost(leader, mesosID, privateIP)
+			if err != nil {
+				return err
+			}
+
+			clientOpts.Proxy, err = detectProxy(ctx, masterProxy, proxyIP)
+			if err != nil {
+				return err
+			}
+
+			sshClient, err := sshclient.NewClient(clientOpts, pluginutil.Logger())
+			if err != nil {
+				return err
+			}
+
+			return sshClient.Run(args)
+		},
+	}
+
+	defUser := "core"
+	cluster, err := ctx.Cluster()
+	if err == nil {
+		if dcosConfUser, ok := cluster.Config().Get("core.ssh_user").(string); ok {
+			defUser = dcosConfUser
+		}
+	}
+	cmd.Flags().BoolVar(&leader, "leader", false, "SSH into the leading master")
+	cmd.Flags().BoolVar(&masterProxy, "master-proxy", false, "Proxy the SSH connection through a master node")
+	cmd.Flags().StringVar(&mesosID, "mesos-id", "", "The agent ID of a node")
+	cmd.Flags().StringVar(&proxyIP, "proxy-ip", "", "Proxy the SSH connection through a different IP address")
+	cmd.Flags().StringVar(&privateIP, "private-ip", "", "Agent node with the provided private IP")
+	cmd.Flags().StringVar(&clientOpts.User, "user", defUser, "The SSH user")
+	cmd.Flags().StringVar(&clientOpts.Config, "config-file", "", "Path to SSH configuration file")
+	cmd.Flags().StringArrayVar(&clientOpts.SSHOptions, "option", nil, "The SSH options")
+
+	return cmd
+}
+
+func detectHost(leader bool, mesosID, privateIP string) (string, error) {
+	if privateIP != "" {
+		return privateIP, nil
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/cmd/node/node_ssh.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	for _, agent := range state.Slaves {
+		if mesosID == agent.ID {
+			host = agent.IP()
+		}
+	}
+	if host == "" {
+		return "", fmt.Errorf("agent '%s' not found", mesosID)
+	}
+	return host, nil
```
bamarni: You could return early here : 
``` go
	for _, agent := range state.Slaves {
		if mesosID == agent.ID {
			return agent.IP(), nil
		}
	}
	return "", fmt.Errorf("agent '%s' not found", mesosID)

```

pkg/sshclient/client.go
```diff
@@ -0,0 +1,88 @@
...
+type Client struct {
+	BinaryPath string
+	Args       []string
+	cmd        *exec.Cmd
+	Logger     *logrus.Logger
+}
+
+var baseArgs = []string{
+	"-A",
+	"-t",
```
bamarni, line 36: Could you add a small comment here stating what these options are for?

pkg/sshclient/client.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/dcos/dcos-cli/api"
+	"github.com/sirupsen/logrus"
+)
+
+// Client to execute commands over SSH.
+type Client struct {
+	BinaryPath string
+	Args       []string
+	cmd        *exec.Cmd
```
bamarni: Seems unused

pkg/cmd/node/node_ssh.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+	}
+	cmd.Flags().BoolVar(&leader, "leader", false, "SSH into the leading master")
+	cmd.Flags().BoolVar(&masterProxy, "master-proxy", false, "Proxy the SSH connection through a master node")
+	cmd.Flags().StringVar(&mesosID, "mesos-id", "", "The agent ID of a node")
+	cmd.Flags().StringVar(&proxyIP, "proxy-ip", "", "Proxy the SSH connection through a different IP address")
+	cmd.Flags().StringVar(&privateIP, "private-ip", "", "Agent node with the provided private IP")
+	cmd.Flags().StringVar(&user, "user", defUser, "The SSH user")
+	cmd.Flags().StringVar(&config, "config-file", "", "Path to SSH configuration file")
+	cmd.Flags().StringSliceVar(&options, "option", nil, "The SSH options. For more information, enter `man ssh_config` in your terminal")
```
bamarni: `man ssh_config` is UNIX specific, I think we should either remove it or append this extra hint based on `runtime.GOOS`

pkg/cmd/node/node_ssh.go
```diff
@@ -0,0 +1,134 @@
...
+			`If your are running this command from another network than DC/OS,
+			consider using "--master-proxy" or "--proxy-ip"`,
+		)
+		return "", nil
+	}
+
+	// masterProxy is true.
+	metadata, err := dcos.NewClient(pluginutil.HTTPClient("")).Metadata()
+	if metadata.PublicIPv4 == "" {
+		return "", fmt.Errorf(`cannot use "--master-proxy", failed to public IP for the master`)
```
bamarni: `failed to detect`?

pkg/sshclient/client.go
```diff
@@ -0,0 +1,88 @@
...
+	c.configureUser(config, user)
+	c.Args = append(c.Args, fmt.Sprintf("%s", host))
+}
+
+// Run adds the optional remote command and starts the SSH session.
+func (c *Client) Run(ctx api.Context, command []string) error {
+	args := c.Args
+	if len(command) > 0 {
+		args = append(args, command...)
+	}
```
bamarni: `args := append(c.Args, command...)` should work fine even when there is no command.

pkg/sshclient/client.go
```diff
@@ -0,0 +1,88 @@
...
+
+var baseArgs = []string{
+	"-A",
+	"-t",
+}
+
+// NewClient creates a new client to start a SSH session.
+func NewClient(options []string, config, user, proxy, host string) (*Client, error) {
+	c := &Client{}
+	c.Logger = pluginutil.Logger()
```
bamarni: Can we pass this as an argument instead? This would make it more easily testable and remove the global dependency.

pkg/sshclient/client.go
```diff
@@ -0,0 +1,88 @@
...
+	args := c.Args
+	if len(command) > 0 {
+		args = append(args, command...)
+	}
+	cmd := exec.Command(c.BinaryPath, args...)
+	c.Logger.Debugf("Running: %v\n", cmd.Args)
+
+	cmd.Stdin = ctx.Input()
+	cmd.Stdout = ctx.Out()
+	cmd.Stderr = ctx.ErrOut()
```
bamarni: it'd be more easily testable if these were passed individually, maybe in the constructor, instead of passing the whole context.

pkg/sshclient/client.go
```diff
@@ -0,0 +1,88 @@
...
+var baseArgs = []string{
+	"-A",
+	"-t",
+}
+
+// NewClient creates a new client to start a SSH session.
+func NewClient(options []string, config, user, proxy, host string) (*Client, error) {
+	c := &Client{}
+	c.Logger = pluginutil.Logger()
+	binPath, err := exec.LookPath("ssh")
```
bamarni: If the binary path was configurable yet defaulted to `exec.LookPath("ssh")`, it'd be easier to test the client, for example by having a fake SSH binary which would make assertions on how it is invoked.

pkg/cmd/node/node_ssh.go
```diff
@@ -0,0 +1,142 @@
...
+			if err != nil {
+				return err
+			}
+
+			sshClient, err := sshclient.NewClient(
+				sshclient.ClientOpts{
+					Input:      ctx.Input(),
+					Out:        ctx.Out(),
+					ErrOut:     ctx.ErrOut(),
+					SSHOptions: options,
```
bamarni: maybe declare the ClientOpts variable before the command? then we don't have to add extra variables for flag parsing : 

    cmd.Flags().StringArrayVar(&opts.SSHOptions, "option", nil, "The SSH options")

bamarni (approved):

PR #180, opened 2019-02-27
Add unit tests on pkg/mesos

rgo3:


pkg/mesos/client_test.go
```diff
@@ -0,0 +1,187 @@
...
+						Hostname: "10.0.0.23",
+					},
+				},
+			},
+		},
+	}
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "/mesos/api/v1", r.URL.String())
+		assert.Equal(t, "application/x-protobuf", r.Header.Get("Accept"))
```
bamarni, line 151: :+1: 

Maybe also `assert.Equal(t, "application/x-protobuf", r.Header.Get("Content-Type"))`

pkg/mesos/client_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "/mesos/api/v1", r.URL.String())
+		var expectedBuffer bytes.Buffer
+		err := json.NewEncoder(&expectedBuffer).Encode(expectedBody)
+		assert.NoError(t, err)
+		requestBody, err := ioutil.ReadAll(r.Body)
+		assert.NoError(t, err)
+		requestBuffer := bytes.NewBuffer(requestBody)
+		assert.Equal(t, &expectedBuffer, requestBuffer)
```
bamarni: maybe it'd have been simpler to just decode the request body and compare the 2 structs, but as you already have it like that we can keep it as is.

pkg/mesos/client_test.go
```diff
@@ -0,0 +1,187 @@
...
+	const expectedAgentID = "9001"
+	expectedBody := master.Call{
+		Type: master.Call_MARK_AGENT_GONE,
+		MarkAgentGone: &master.Call_MarkAgentGone{
+			AgentID: mesos.AgentID{Value: expectedAgentID},
+		},
+	}
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "/mesos/api/v1", r.URL.String())
```
bamarni, line 176: could be nice to also assert on the request method being `POST` here, and asserting on `GET` in other tests.

armandgrillet (approved):

armandgrillet:
`ok  	github.com/dcos/dcos-core-cli/pkg/mesos	1.169s	coverage: 74.1% of statements`
Nice!

bamarni (approved):
Awesome, we're getting so close to our 75% coverage threshold :smile:

PR #203, opened 2019-03-22
Support core.mesos_master_url

rgo3:
https://jira.mesosphere.com/browse/DCOS-49238

### DIFF
--- a/pkg/cmd/node/node.go
+++ b/pkg/cmd/node/node.go
@@ -49,6 +49,20 @@ func diagnosticsClient() *diagnostics.Client {
 	return diagnostics.NewClient(pluginutil.HTTPClient(""))
 }
 
-func mesosClient() *mesos.Client {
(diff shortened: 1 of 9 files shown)


pkg/cmd/node/node.go
```diff
@@ -49,6 +49,18 @@ func diagnosticsClient() *diagnostics.Client {
 	return diagnostics.NewClient(pluginutil.HTTPClient(""))
 }
 
-func mesosClient() *mesos.Client {
+func mesosDNSClient() *mesos.Client {
```
armandgrillet, line 53: I don't think `mesosDNSClient` is the right name for this as it does not return a different client class. `mesosDefaultClient` would be clearer IMO.

armandgrillet (changes requested):

armandgrillet (approved):

PR #207, opened 2019-03-28
Specified if agent is public when using 'dcos node list'.

armandgrillet:


pkg/cmd/node/node_list.go
```diff
@@ -108,7 +108,12 @@ func newCmdNodeList(ctx api.Context) *cobra.Command {
 
 			slaves := state.Slaves
 			for _, s := range slaves {
-				s.Type = "agent"
```
rgo3: If you leave that line there you can get rid of the else statement.

rgo3: The same pattern is used when iterating over the masters.

PR #215, opened 2019-04-05
Add journlctl format options: `json-pretty`, `json`, `cat`, `short``

rgo3:
https://jira.mesosphere.com/browse/DCOS_OSS-4853

### DIFF
--- a/pkg/logs/client.go
+++ b/pkg/logs/client.go
@@ -22,6 +22,14 @@ type Client struct {
 	colored bool
 }
 
+// Options encapsulates options that can be set via flags on the command.
+type Options struct {
+	Filters []string
+	Follow  bool
+	Format  string
+	Skip    int
+}
+
 // NewClient creates a new logs client.
 func NewClient(baseClient *httpclient.Client, out io.Writer) *Client {
 	c := &Client{http: baseClient, out: out}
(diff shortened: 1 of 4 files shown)


rgo3:

pkg/logs/client_test.go
```diff
@@ -89,7 +89,12 @@ func TestPrintComponent(t *testing.T) {
 		c := NewClient(pluginutil.HTTPClient(ts.URL), &b)
 		c.colored = fixture.colored
 
-		err := c.PrintComponent(fixture.route, fixture.service, fixture.skip, fixture.filters, false)
+		opts := Options{
+			Skip:    fixture.skip,
+			Filters: fixture.filters,
+		}
+
+		err := c.PrintComponent(fixture.route, fixture.service, opts)
```
bamarni, line 121: Could we have tests for other formats than `short`?

pkg/logs/client.go
```diff
@@ -74,24 +82,58 @@ func (c *Client) PrintComponent(route string, service string, skip int, filters
...
+	switch opts.Format {
+	case "json-pretty":
+		enc.SetIndent("", "    ")
+		fallthrough
+	case "json":
+		return enc.Encode(entry.Fields)
+	case "cat":
+		c.colorPrefix(entry.Fields.Priority)
+		fmt.Fprint(c.out, entry.Fields.Message)
+		c.colorPostfix()
```
bamarni: Did you mean `colorSuffix`? Also, this function is actually resetting the attributes, maybe it can be called `ansiReset` or `resetColor`.

As a side note, the way we're dealing with colors here is very manual, in the future we should consider having a more intuitive color API / library to use in the CLI.

bamarni (approved):

armandgrillet:

PR #249, opened 2019-05-15
[1.13] Add completion for `dcos node log --component=`

bamarni:
This also renames `dcos node list-components` to `dcos node list-units`
as it's listing DC/OS systemd units. There is a new `--type` option to
specify which type of units to list.


armandgrillet:
Could we add a deprecation message for `dcos node list-components`?

rgo3 (approved):

PR #252, opened 2019-05-20
Added 'dcos task list'.

armandgrillet:


pkg/cmd/task/task_list.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+
+			tableHeader := []string{"NAME", "HOST", "USER", "STATE", "ID", "AGENT ID", "REGION", "ZONE"}
+			table := cli.NewTable(ctx.Out(), tableHeader)
+
+			if quietOutput {
+				for _, t := range tasks {
+					fmt.Fprintln(ctx.Out(), t.Name)
+				}
+			}
```
bamarni, line 54: there should be a return here, also it could be moved before creating the table

pkg/cmd/task/task_list.go
```diff
@@ -0,0 +1,136 @@
...
+		switch completed {
+		case true:
+			if completedTaskStates[t.State.String()] {
+				result = append(result, t)
+			}
+		case false:
+			if !completedTaskStates[t.State.String()] {
+				result = append(result, t)
+			}
+		}
```
bamarni, line 127:
``` go
for _, t := range tasks {
	if completed == completedTaskStates[t.State.String()] {
		result = append(result, t)
	}
}
```

pkg/cmd/task/task_list.go
```diff
@@ -0,0 +1,136 @@
...
+		"TASK_FINISHED":         true,
+		"TASK_KILLED":           true,
+		"TASK_FAILED":           true,
+		"TASK_LOST":             true,
+		"TASK_ERROR":            true,
+		"TASK_GONE":             true,
+		"TASK_GONE_BY_OPERATOR": true,
+		"TASK_DROPPED":          true,
+		"TASK_UNREACHABLE":      true,
+		"TASK_UNKNOWN":          true,
```
bamarni: Are there constants for this in the mesos-go project?

pkg/cmd/task/task_list.go
```diff
@@ -0,0 +1,136 @@
...
+					if f.FrameworkInfo.ID.GetValue() == t.FrameworkID.Value {
+						user = f.FrameworkInfo.User
+					}
+				}
+
+				item := []string{
+					t.Name,
+					host,
+					user,
+					strings.Replace(t.State.String(), "TASK_", "", 1),
```
bamarni: nit: `TrimPrefix`

bamarni:
did you try to run integration tests against these changes?

armandgrillet:
We cannot run the integration tests until the entire subcommand is in Go.

pkg/cmd/task/task_list.go
armandgrillet: Yes but they're not assembled in something similar to `completedTaskStates`.

bamarni: We could still benefit from them to avoid hardcoding strings, eg. `mesos.TASK_UNKNOWN: true,`

PR #254, opened 2019-05-21
Added 'dcos task ls'.

armandgrillet:


pkg/cmd/task/task_ls.go
```diff
@@ -0,0 +1,129 @@
...
+			}
+
+			allTasks, err := mesosClient.Tasks()
+			if err != nil {
+				return err
+			}
+
+			var tasks []lib.Task
+			for _, t := range allTasks {
+				if strings.Contains(t.TaskID.Value, args[0]) {
```
bamarni: Can we have the same task ID matching logic than the current CLI?

> A full task ID, a partial task ID, or a Unix shell wildcard pattern (eg. 'my-task*').

pkg/cmd/task/task_ls.go
```diff
@@ -0,0 +1,129 @@
...
+				}
+			}
+
+			for i, t := range tasks {
+				if len(tasks) > 1 {
+					fmt.Fprintln(ctx.Out(), "===> "+t.TaskID.Value+" <===")
+				}
+				taskPath := "/frameworks/" + t.FrameworkID.Value + "/executors/" + executorIDs[i] + "/runs/" + containerParentIDs[i]
+				for key, value := range agentsPaths[t.AgentID.Value] {
+					if strings.Contains(key, taskPath) {
```
bamarni: Should this be `HasPrefix`?

pkg/cmd/task/task_ls.go
```diff
@@ -0,0 +1,131 @@
...
+				}
+			}
+
+			for i, t := range tasks {
+				if len(tasks) > 1 {
+					fmt.Fprintln(ctx.Out(), "===> "+t.TaskID.Value+" <===")
+				}
+				taskPath := "/frameworks/" + t.FrameworkID.Value + "/executors/" + executorIDs[i] + "/runs/" + containerParentIDs[i]
+				for key, value := range agentsPaths[t.AgentID.Value] {
+					if strings.HasSuffix(key, taskPath) {
```
bamarni, line 93: Should it really be suffix and not prefix?

armandgrillet: Yes, I've manually tested and the function to use in order to get the same behavior as in the Python CLI is to use `HasSuffix`.

bamarni (approved):

PR #264, opened 2019-06-13
Port the `dcos task attach/exec` commands to Go

bamarni:
Checklist:

- [x] support `core.mesos_master_url`
- [x] `dcos task attach` (https://jira.mesosphere.com/browse/DCOS_OSS-4794)
- [x] `dcos task exec` (https://jira.mesosphere.com/browse/DCOS_OSS-4795)
- [x] support `dcos task exec --user` (https://jira.mesosphere.com/browse/DCOS-44368)
- [x] `dcos task exec` verbose output should give the container id of the created task (https://jira.mesosphere.com/browse/DCOS_OSS-4728)
- [ ] add tests
- [x] maybe catch SIGINT and SIGTERM signals and close the HTTP connections when encountered. This is especially important for the `ATTACH_CONTAINER_INPUT` call, if the client process terminates without closing the HTTP connection properly, Mesos will think we're still attached to the input and will never clean-up the connection (see https://issues.apache.org/jira/browse/MESOS-9838). Further calls to `dcos task attach` will then fail with `Multiple input connections are not allowed`.
- [ ] Make the `WAIT_CONTAINER` call concurrently with the others (https://jira.mesosphere.com/browse/DCOS_OSS-5282)


design/attach.md
```diff
@@ -0,0 +1,41 @@
+# attach
+
+The `dcos task attach` command can attach terminal standard streams to a running container.
```
armandgrillet: s/can attach terminal standard streams/attaches the terminal standard streams

design/attach.md
```diff
@@ -0,0 +1,41 @@
...
+
+It is then able to stream the STDIN data of the CLI process over to the container, as well as
```
armandgrillet: I think we can remove the word "data" here and for STDOUT/STDERR.

design/attach.md
```diff
@@ -0,0 +1,41 @@
...
+streaming the STDOUT/STDERR data of the container back to the CLI process.
+
+The command only works for tasks that were launched with with the Universal Container Runtime (UCR)
```
armandgrillet: Remove one "with".

design/attach.md
```diff
@@ -0,0 +1,41 @@
...
+and with a TTY.
+
+## Usage
+
+    dcos task attach [--no-stdin] <task-id>
+
+The `<task-id>` mandatory argument indicates the task to attach to. It can be a full task ID,
```
armandgrillet: s/The `<task-id>` mandatory argument/The  mandatory argument `<task-id>`

armandgrillet: s/indicates the task/indicates which task

pkg/cmd/task/task.go
```diff
@@ -32,9 +40,53 @@ func NewCommand(ctx api.Context) *cobra.Command {
...
+	}
+
+	allTasks, err := mesosClient.Tasks()
+	if err != nil {
+		return nil, err
+	}
+
+	var tasks []mesosgo.Task
+	for _, t := range allTasks {
+		if strings.Contains(t.TaskID.Value, id) {
```
armandgrillet: This doesn't go along the design doc: "It then loops through the tasks and tries to find a match which is exactly `<task-id>`, contains`<task-id>`, or matches the wildcard pattern of `<task-id>`."

pkg/cmd/task/task.go
```diff
@@ -32,9 +40,53 @@ func NewCommand(ctx api.Context) *cobra.Command {
...
+
+	cluster, err := ctx.Cluster()
+	if err != nil {
+		return nil, err
+	}
+
+	httpClient := httpcli.New(
+		httpcli.Endpoint(fmt.Sprintf("%s/slave/%s/api/v1", cluster.URL(), agentID)),
+		httpcli.Do(httpcli.With(httpcli.RoundTripper(rt))),
+		httpcli.Codec(codecs.ByMediaType[codecs.MediaTypeJSON]),
```
armandgrillet, line 143: A comment for this line would be good.

pkg/mesos/taskio.go
```diff
@@ -0,0 +1,343 @@
...
+			defer cancel()
+
+			err := t.attachContainerInput(ctx)
+			if err != nil {
+				return
+			}
+		}()
+	}
+
+	// Attach container outputs to our terminal stdout/stderr.
```
armandgrillet: s/our/the

armandgrillet:
```
pkg/mesos/taskio.go:24:6: exported type TaskIOOpts should have comment or be unexported
pkg/mesos/taskio.go:35:6: exported type TaskIO should have comment or be unexported
pkg/mesos/taskio.go:40:1: exported function NewTaskIO should have comment or be unexported
pkg/mesos/taskio.go:62:1: exported method TaskIO.Attach should have comment or be unexported
```

pkg/cmd/task/task.go
```diff
@@ -32,9 +41,104 @@ func NewCommand(ctx api.Context) *cobra.Command {
...
+		agents, err := mesosClient.Agents()
+		if err != nil {
+			return nil, err
+		}
+
+		for _, a := range agents {
+			if a.AgentInfo.ID.Value == agentID {
+				// PID format: Host@IP:Port
+				s1 := strings.Split(a.GetPID(), "@")
+				s2 := strings.Split(s1[1], ":")
```
jdef: splitting s1 to arrive at s2, and then re-joining it here seems strange. why not just use `"http://"+s1[1]` ?

pkg/cmd/task/task.go
```diff
@@ -32,9 +41,104 @@ func NewCommand(ctx api.Context) *cobra.Command {
...
+				url := fmt.Sprintf("http://%s:%s", s2[0], s2[1])
+
+				return httpcli.New(
+					httpcli.Endpoint(url),
+
+					// Use JSON over protobuf to ease logging.
+					httpcli.Codec(codecs.ByMediaType[codecs.MediaTypeJSON]),
```
jdef, line 129: JSON has problems w/ unknown enums. you're usually much better off (in production code) using protobuf

pkg/cmd/task/task.go
```diff
@@ -32,9 +41,104 @@ func NewCommand(ctx api.Context) *cobra.Command {
...
+
+		agents, err := mesosClient.Agents()
+		if err != nil {
+			return nil, err
+		}
+
+		for _, a := range agents {
+			if a.AgentInfo.ID.Value == agentID {
+				// PID format: Host@IP:Port
+				s1 := strings.Split(a.GetPID(), "@")
```
jdef: why bother with PID, AgentInfo has a `hostname` and `port` field

pkg/mesos/taskio.go
```diff
@@ -0,0 +1,370 @@
...
+				if err := forward(data.GetData(), t.opts.Stderr); err != nil {
+					return err
+				}
+			}
+		}
+	}
+}
+
+// waitContainer waits for the container to terminate and returns its exit code.
+func (t *TaskIO) waitContainer() (int, error) {
```
jdef, line 429: waitContainer should probably take a context parameter

pkg/mesos/taskio.go
```diff
@@ -0,0 +1,370 @@
...
+	// We are only able to get the 'exit_status' of tasks launched via the default executor
+	// (i.e. as pods rather than via the command executor). In the future, mesos will deprecate
+	// the command executor in favor of the default executor, so this check will  be able to go away.
+	// In the meantime, we will always return '0' for tasks launched via the command executor.
+	if t.containerID.GetParent() == nil {
+		return 0, nil
+	}
+
+	call := agentcalls.NonStreaming(agentcalls.WaitContainer(t.containerID))
+	resp, err := t.opts.Sender.Send(context.TODO(), call)
```
jdef: should pass in a real context here (probably the one that I suggested as a param for waitContainer)

pkg/cmd/task/task.go
bamarni: I ended-up copying the logic of our Python codebase, as we know it is working.

I've added a TODO comment as it requires investigation to make sure the fields are equivalent and always there.

pkg/cmd/task/task.go
bamarni: This is mainly for debugging purpose, when users run the CLI in verbose mode they can then see HTTP request/response JSON bodies being printed, this is not possible with binary payloads.

Also, the Python codebase currently uses JSON as well. Do you have concrete examples of a call which wouldn't work?

pkg/mesos/taskio.go
bamarni: I've added a context with timeout, and comments about a current limitation, see https://mesosphere.slack.com/archives/C02CGA57K/p1560869602063100

pkg/cmd/task/task.go
armandgrillet: We had the issue once where Mesos added a key in one of its response and the CLI started to break, it was solved here: https://github.com/dcos/dcos-core-cli/commit/8017331eb64fba73bbca498569a768ae730288c9#diff-65145422e82c3e8089c40d2b75584a43
The Python codebase is not affected AFAIK because the issue happens when we are decoding the JSON response into a mesos-go struct.

pkg/cmd/task/task_exec.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+func newCmdTaskExec(ctx api.Context) *cobra.Command {
+	var interactive, tty bool
+	var user string
+
+	cmd := &cobra.Command{
+		Use:   "exec <task> <cmd> [<args>...]",
+		Short: "Launch a process (<cmd>) inside of a container for a task (<task>).",
+		Args:  cobra.MinimumNArgs(2),
```
bamarni, line 20: TODO: we should disable flag parsing after the `<task>` argument, not to have cobra fail or mix flags of the exec command with potential flags of the command to exec.

armandgrillet: https://jira.mesosphere.com/browse/DCOS_OSS-5283

armandgrillet (approved):

rgo3 (approved):

pkg/cmd/task/task.go
```diff
@@ -115,10 +115,11 @@ func mesosHTTPClient(ctx api.Context, agentID string) (*httpcli.Client, error) {
...
 		for _, a := range agents {
 			if a.AgentInfo.ID.Value == agentID {
+
 				// PID format: Host@IP:Port
-				s1 := strings.Split(a.GetPID(), "@")
-				s2 := strings.Split(s1[1], ":")
-				url := fmt.Sprintf("http://%s:%s", s2[0], s2[1])
+				// TODO: AgentInfo provides a hostname and a port, investigate if they can be used instead.
+				ipPort := strings.Split(a.GetPID(), "@")
+				url := fmt.Sprintf("http://%s", ipPort)
```
jdef, line 123: `http` probably won't work in strict mode clusters, you'll want `https` for that, right?

bamarni: This code block is getting a client for pure mesos clusters. Users (usually core team members), can do eg. `dcos config set core.mesos_master_url http://localhost:9090`, and then the CLI will use this cluster for Mesos related tasks. The URL creation logic was taken from our Python codebase: https://github.com/dcos/dcos-core-cli/blob/90c3bf3e1b3198c6ed06242cebd2998e1ea67b06/python/lib/dcos/dcos/mesos.py#L662

For DC/OS cluster it happens a few lines below, where we create a Mesos HTTP client with our custom RoundTripper, which is handling HTTPS if the cluster was setup accordingly when doing `dcos cluster setup`.

bamarni:
We're going to merge it to close the sprint, the command is not enabled for now, we have a few follow-up tickets for the next sprint (eg. adding tests) before we can enable the command.

PR #265, opened 2019-06-13
Allows printing logs of multiple tasks.

armandgrillet:
This does not accept `--follow` yet.

### DIFF
--- a/pkg/cmd/task/task_log.go
+++ b/pkg/cmd/task/task_log.go
@@ -49,21 +49,34 @@ func newCmdTaskLog(ctx api.Context) *cobra.Command {
 
 			if len(tasks) == 0 {
 				return fmt.Errorf("no task ID found containing '%s'", args[0])
-			} else if len(tasks) > 1 {
-				return fmt.Errorf("found more than one task with the same name: %v", tasks)
 			}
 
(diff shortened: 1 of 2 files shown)


bamarni:

pkg/cmd/task/task_log.go
```diff
@@ -49,21 +49,34 @@ func newCmdTaskLog(ctx api.Context) *cobra.Command {
...
+						ctx.Logger().Info(`Task logs don't support output options. Defaulting to "short"...`)
+					}
+					opts := logs.Options{
+						Follow: follow,
+						Format: output,
+						Skip:   -1 * lines,
+					}
+					err = logClient.PrintTask(task, file, opts)
+					if err != nil {
+						return err
```
rgo3, line 72: Instead of immediately returning we could only log the error here so that we potentially still get logs from other tasks. At least we decided to do that in other places of the CLI.

rgo3 (approved):
Otherwise LGTM

pkg/cmd/task/task_log.go
armandgrillet: OK, will change that when working on following multiple tasks.

PR #269, opened 2019-06-19
Add 'dcos task download' command

rgo3:


pkg/cmd/task/task.go
```diff
@@ -46,6 +46,7 @@ func NewCommand(ctx api.Context) *cobra.Command {
 		newCmdTaskList(ctx),
 		newCmdTaskLog(ctx),
 		newCmdTaskLs(ctx),
+		newCmdTaskDownload(ctx),
```
armandgrillet: Should be moved right after `newCmdTaskAttach(ctx),`.

pkg/cmd/task/task_download.go
```diff
@@ -0,0 +1,174 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/dcos/dcos-cli/api"
+	"github.com/dcos/dcos-core-cli/pkg/mesos"
+	"github.com/dcos/dcos-core-cli/pkg/pluginutil"
+	"github.com/spf13/cobra"
+)
+
+const description = `Downloads files from the sandbox of a given task.
```
armandgrillet, line 15: I don't think we commonly use a `const` for the description, was this done because it is a longer piece of text than usual?

pkg/cmd/task/task_download.go
```diff
@@ -0,0 +1,174 @@
...
+
+	var failed []string
+	for _, file := range files {
+		err := <-errors
+		if err != nil {
+			failed = append(failed, path.Base(file.Path))
+		}
+	}
+	if len(failed) != 0 {
+		return fmt.Errorf("Download of following files failed: [%s]", strings.Join(failed, ", "))
```
armandgrillet: I don't think `%s` needs to be surrounded by square brackets in that error message.

armandgrillet (approved):
Minor issues to fix, otherwise LGTM.

pkg/cmd/task/task_download.go
rgo3: Yes, using it a multiline string directly in the `cmd` struct looked pretty ugly.

pkg/cmd/task/task_download.go
rgo3: This doesn't really work as I want anyway because when I pull an error from the channel it doesn't necessarily belong to the file I am looping over at that moment. Will change it to something different.

pkg/cmd/task/task_download.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/dcos/dcos-core-cli/pkg/mesos"
+	"github.com/dcos/dcos-core-cli/pkg/pluginutil"
+	"github.com/spf13/cobra"
+)
+
+const description = `Downloads files from the sandbox of a given task.
+The <path> argument can specify a pattern for a single or multiple files.
+If no <path> is provided, the entire sandbox will be downloaded.`
+
+// newCmdMesosDownload downloads files from a tasks sandbox.
```
bamarni:
```suggestion
// newCmdTaskDownload downloads files from a tasks sandbox.
```

pkg/cmd/task/task_download.go
```diff
@@ -0,0 +1,174 @@
...
+		Use:   "download <task-id> [<path>]",
+		Short: description,
+		Args:  cobra.RangeArgs(1, 2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			task, err := findTask(ctx, args[0])
+			if err != nil {
+				return err
+			}
+
+			lastStatus := task.Statuses[0]
```
bamarni:
```suggestion
			firstStatus := task.Statuses[0]
```

pkg/cmd/task/task_download.go
```diff
@@ -0,0 +1,174 @@
...
+			containerID := lastStatus.ContainerStatus.GetContainerID().Value
```
bamarni:
```suggestion
			containerID := firstStatus.ContainerStatus.GetContainerID().Value
```

pkg/cmd/task/task_download.go
```diff
@@ -0,0 +1,174 @@
...
+
+			executorID := task.TaskID.Value
+			if task.ExecutorID != nil {
+				executorID = (*task.ExecutorID).Value
```
bamarni: why dereferencing here? does `executorID = task.ExecutorID.Value` work?

pkg/cmd/task/task_download.go
```diff
@@ -0,0 +1,174 @@
...
+func getPath(paths map[string]string, framework, executor, container string) (string, error) {
+	for k, v := range paths {
+		if strings.Contains(k, "/frameworks/"+framework+"/executors/"+executor+"/runs/"+container) {
+			return v, nil
+		}
+	}
+	return "", fmt.Errorf("unable to find task")
+}
+
+func downloadDir(c *mesos.Client, ctx api.Context, agentID, agentPath, targetDir, pattern string) error {
```
bamarni: maybe just `download`? As it can also be a file

pkg/cmd/task/task_download.go
```diff
@@ -0,0 +1,174 @@
...
+
+	if strings.HasPrefix(parent, "/") {
+		return agentPath + parent
+	}
+
+	return agentPath + "/" + parent
+}
+
+func basePath(pattern string) string {
+	base := path.Base(pattern)
```
bamarni:
```suggestion
	base := filepath.Base(pattern)
```

pkg/cmd/task/task_download.go
```diff
@@ -0,0 +1,174 @@
...
+	parent := filepath.Dir(pattern)
+	if parent == "/" || parent == "." {
+		return agentPath
+	}
+
+	if strings.HasPrefix(parent, "/") {
+		return agentPath + parent
+	}
+
+	return agentPath + "/" + parent
```
bamarni: It should be possible to use `filepath.Join(agentPath, parent)` and it would take care of separators normally.

pkg/cmd/task/task_download.go
```diff
@@ -0,0 +1,174 @@
...
+		if err != nil {
+			return fmt.Errorf("'%s' as pattern not supported: %v", pattern, err)
+		}
+
+		if !matched {
+			errors <- nil
+			continue
+		}
+
+		go func(f mesos.File) {
```
bamarni, line 103: maybe the content of the function could be extracted to a separate function for readability

rgo3: I first planned on doing that but the extraction created a function with a really long list or arguments to pass, so this looked cleaner to me.

pkg/cmd/task/task_download.go
rgo3: Just `status` should suffice.

pkg/cmd/task/task_download.go
bamarni: We could consider making a dedicated struct for this at some point.

bamarni (approved):

PR #275, opened 2019-06-27
Enables task log --follow on multiple tasks.

armandgrillet:
Tested by running this taks on a cluster: 

```
{
  "id": "/test-service",
  "backoffFactor": 1.15,
  "backoffSeconds": 1,
  "cmd": "while true; do hostname && sleep 120; done",
  "container": {
    "type": "DOCKER",
    "volumes": [],
    "docker": {
      "image": "ubuntu",
      "forcePullImage": false,
      "privileged": false,
      "parameters": []
    }
  },
  "cpus": 0.1,
  "disk": 0,
  "instances": 3,
  "mem": 128,
  "networks": [
    {
      "mode": "host"
    }
  ]
}
```

And ten running `build/darwin/dcos task log test-service --follow`.

### DIFF
--- a/pkg/cmd/task/task_log.go
+++ b/pkg/cmd/task/task_log.go
@@ -35,30 +35,76 @@ func newCmdTaskLog(ctx api.Context) *cobra.Command {
 
 			// Only one task matching or multiple tasks but no follow.
 			if len(tasks) == 1 || follow == false {
+				if output != "short" {
+					output = "short"
+					ctx.Logger().Info(`Task logs don't support output options. Defaulting to "short"...`)
+				}
+
+				var failed bool
 				for _, task := range tasks {
 					if len(tasks) > 1 {
 						fmt.Fprintln(ctx.Out(), fmt.Sprintf("===> %s <===", task.TaskID.Value))
 					}
 					logClient := logs.NewClient(pluginutil.HTTPClient(""), ctx.Out())
-					if output != "short" {
-						output = "short"
-						ctx.Logger().Info(`Task logs don't support output options. Defaulting to "short"...`)
-					}
 					opts := logs.Options{
 						Follow: follow,
 						Format: output,
 						Skip:   -1 * lines,
 					}
-					return logClient.PrintTask(task.TaskID.Value, file, opts)
+					err := logClient.PrintTask(task.TaskID.Value, file, opts)
+					if err != nil {
+						failed = true
+						fmt.Fprintf(ctx.ErrOut(), "Error: %v\n", err)
+					}
 				}
+
+				if len(tasks) > 1 && failed {
+					return fmt.Errorf("could not log all matched tasks")
+				}
+				return nil
 			}
 
-			// TODO (DCOS_OSS-5153): multiple followed tasks.
-			var taskNames []string
-			for _, task := range tasks {
-				taskNames = append(taskNames, task.TaskID.Value)
+			// Follow multiple tasks.
+			if output != "cat" {
+				ctx.Logger().Info(`Task logs don't support output options. Defaulting to "cat"...`)
 			}
-			return fmt.Errorf("found more than one task with the same name, unable to follow them all: %v", taskNames)
+
+			// The channel receiving the content of the logs. Each task followed dumps its logs in it.
+			msgChan := make(chan taskData)
+			errChan := make(chan error)
+			for _, task := range tasks {
+				go func(taskID string, lines int, file string, c chan taskData, e chan error) {
+					taskOut := &taskWriter{
+						task:   taskID,
+						writer: c,
+					}
+					logClient := logs.NewClient(pluginutil.HTTPClient(""), taskOut)
+					opts := logs.Options{
+						Follow: true,
+						Format: "cat",
+						Skip:   -1 * lines,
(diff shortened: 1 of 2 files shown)


pkg/cmd/task/task_log.go
```diff
@@ -51,14 +51,50 @@ func newCmdTaskLog(ctx api.Context) *cobra.Command {
 					}
 					return logClient.PrintTask(task.TaskID.Value, file, opts)
 				}
-			}
+			} else {
+				msgChan := make(chan taskMessage)
+				errChan := make(chan error)
+
+				for _, task := range tasks {
```
rgo3: Is it possible to have the closure in a separate function? In general we might want to short this `if {...} else {...}` block by sourcing some code out to functions.

pkg/cmd/task/task_log.go
```diff
@@ -51,14 +51,50 @@ func newCmdTaskLog(ctx api.Context) *cobra.Command {
 					}
 					return logClient.PrintTask(task.TaskID.Value, file, opts)
```
rgo3: Even though this is not part of the PR it caught my eye, how do we print multiple if we return after the first one?

rgo3 (changes requested):

pkg/cmd/task/task_log.go
bamarni: The follow case seems to be on the else statement below

pkg/cmd/task/task_log.go
```diff
@@ -35,30 +35,52 @@ func newCmdTaskLog(ctx api.Context) *cobra.Command {
...
 						Format: output,
 						Skip:   -1 * lines,
 					}
 					return logClient.PrintTask(task.TaskID.Value, file, opts)
 				}
-			}
+			} else {
+				if output != "cat" {
+					output = "cat"
+					ctx.Logger().Info(`Task logs don't support output options. Defaulting to "cat"...`)
```
bamarni: This could be put before the if else statement no to duplicate with the other one.

armandgrillet: They don't have the same output limitation (first one is `short`, the second one is `cat`).

pkg/cmd/task/task_log.go
```diff
@@ -35,30 +35,52 @@ func newCmdTaskLog(ctx api.Context) *cobra.Command {
...
+				errChan := make(chan error)
+				for _, task := range tasks {
+					go followAndlog(task.TaskID.Value, lines, file, msgChan, errChan)
+				}
+
+				lastTask := ""
+				for {
+					select {
+					case newTaskMsg := <-msgChan:
+						if string(newTaskMsg.data) != "" {
```
bamarni: As on line 106 we check for `if len(data) > 1` before sending to this channel, this check seems unnecessary.

pkg/cmd/task/task_log.go
```diff
@@ -68,3 +88,38 @@ func newCmdTaskLog(ctx api.Context) *cobra.Command {
...
+	data string
+}
+
+type taskWriter struct {
+	task     string
+	receiver chan taskMessage
+}
+
+func (t *taskWriter) Write(data []byte) (n int, err error) {
+	// When the data is only one byte, it doesn't represent a message.
```
bamarni: what are the 1 byte data that we can receive here?

pkg/cmd/task/task_log.go
bamarni: Ah, I think Robin's point is that we're on a loop here.

pkg/logs/client.go
```diff
@@ -144,6 +123,50 @@ func (c *Client) PrintTask(taskID string, file string, opts Options) error {
...
+		return err
+	}
+	defer client.Unsubscribe(events)
+
+	for msg := range events {
+		if len(msg.Data) == 0 {
+			continue
+		}
+
+		if print {
```
bamarni: ideally we don't overwrite the built-in print function

pkg/logs/client.go
```diff
@@ -144,6 +123,50 @@ func (c *Client) PrintTask(taskID string, file string, opts Options) error {
...
+	return nil
+}
+
+func (c *Client) dumpEntry(rawEntry []byte) error {
+	var entry Entry
+	err := json.Unmarshal(rawEntry, &entry)
+	if err != nil {
+		return err
+	}
+	fmt.Fprint(c.out, entry.Fields.Message)
```
bamarni, line 167: For the record I'm not a big fan of this approach, if someone uses the client with `os.Stdout` for example, they would get all the entries next to each other without new lines separator. There should be a well-defined API for following logs, eg. using a channel instead of an io.Writer.

pkg/logs/client.go
```diff
@@ -211,6 +234,9 @@ func (c *Client) resetColor() {
 }
 
 func httpResponseToError(resp *http.Response) error {
+	if resp.StatusCode == 204 {
+		return fmt.Errorf("unexpected status code %d, no logs found", resp.StatusCode)
```
bamarni: Could this just be `no logs found`?

bamarni (changes requested):

PR #278, opened 2019-06-27
Fix possible segfault in `node ssh` subcommand

makkes:
When no metadata can be retrieved for a cluster, `Metadata()` returns
nil and an error so let's check whether an error has been returned.


makkes:
@bamarni @ArmandGrillet @rgo3 would you mind reviewing this?

bamarni (approved):

bamarni:
Thanks Max, LGTM

PR #274, opened 2019-06-26
Enable task Go subcommand in dcos-core-cli

bamarni:
https://jira.mesosphere.com/browse/DCOS_OSS-5248

PR #282, opened 2019-06-28
Send an empty process IO data on STDIN EOF

bamarni:
For an `ATTACH_CONTAINER_INPUT` call and when STDIN reaches EOF,
Mesos expects an empty process IO data message, as highlighted in:

https://github.com/dcos/dcos-core-cli/blob/7b67c5a369ff92d1c7776fa5f2b1560406a6d27d/python/lib/dcos/dcos/mesos.py#L1657-L1661

This is currently not done in the new attach and exec commands in Go. It
causes them to exit with an erroneous exit code in some cases. It can be
seen by reaching EOF while running a cat command for example. The
following command returns with an exit code of 137 instead of 0:

    dcos task exec -i my-app cat <input.txt

However, sending an empty process IO message turned out not to be doable in JSON,
due to mesos-go omitting the field when it's empty:

    {"type":"PROCESS_IO","process_io":{"type":"DATA","data":{"type":"STDIN"}}}

Mesos then returns:

    Error: malformed request: Expecting 'process_io.data.data' to be present

To get around this issue, we're switching to protobuf.

### DIFF
--- a/pkg/cmd/task/task.go
+++ b/pkg/cmd/task/task.go
@@ -10,7 +10,6 @@ import (
 	"github.com/dcos/dcos-core-cli/pkg/pluginutil"
 	"github.com/gobwas/glob"
 	mesosgo "github.com/mesos/mesos-go/api/v1/lib"
-	"github.com/mesos/mesos-go/api/v1/lib/encoding/codecs"
 	"github.com/mesos/mesos-go/api/v1/lib/httpcli"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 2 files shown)


bamarni:
@jdef FYI

bamarni:
This was seen as some integration tests failed while enabling the command in https://github.com/dcos/dcos-core-cli/pull/274

jdef:
> @jdef FYI

I'm not surprised to see yet another problem w/ using JSON as the encoding. Thanks for the heads up.

armandgrillet (approved):

jdef:
Regarding Mesos STDIN expectations, documentation appears to be lacking here https://github.com/apache/mesos/blob/master/docs/operator-http-api.md#attach_container_input

bamarni:
See https://github.com/apache/mesos/pull/340 for the doc update

PR #286, opened 2019-07-04
Fixed dcos task exec with parent container.

armandgrillet:
This adds the parent contained ID to the container ID used by TaskIO, we have previously seen errors in the new Go codebase due to the missing root container ID (see https://jira.mesosphere.com/browse/DCOS-56057).

Tested manually.


armandgrillet:

bamarni:
I think it should also be done in task attach.

As the code is quite similar we could merge it in a `newTaskIO(id string) (*TaskIO, error)` function.

armandgrillet:
The CI failing on Windows is not due to the change, waiting on the macOS integration tests and then will merge. I'll improve the code following Bilal's suggestions in another PR.

bamarni (approved):

PR #287, opened 2019-07-04
Add a newTaskIO helper function

bamarni:
As a follow-up from https://github.com/dcos/dcos-core-cli/pull/286,
we're adding a helper function to reduce code duplication and make sure
the TaskIO created for attach and exec have the same semantics
(containing the parent ID, etc.).


bamarni:

PR #289, opened 2019-07-05
Specified public agent when using 'dcos node list'.

armandgrillet:


armandgrillet:

pkg/cmd/node/node_list.go
```diff
@@ -123,6 +123,11 @@ func newCmdNodeList(ctx api.Context) *cobra.Command {
 
 				nodes = append(nodes, s)
 
+				// Additional information, only for non JSON output.
+				if val, ok := s.Attributes["public_ip"].(string); ok && val == "true" {
+					s.Type = "agent (public)"
```
bamarni, line 128: to be safe, shall we move the JSON block on top of the command to make sure we don't alter stuff there?

armandgrillet: No because we modify a few values, I believe it's fine as it is with comments.

PR #295, opened 2019-07-12
Increase the buffer size when reading from piped stdin

bamarni:
We had setup a 1 byte buffer for reading from stdin to work around an
issue with the escape proxy reader we're using. However @vespian
reported that it affects performances of the `dcos task exec` command
quite badly when the input is piped and has a significant size.

This updates the code to use a 1 byte buffer only for the terminal
use-case, and increasing it to 512 for piped input.

Running the following command with a ~50MB file showed the following
results:

    dcos task exec -i my-test sh -c "cat >/dev/null" < /path/to/big/file

Former Python command: `0.67s user 0.07s system 2% cpu 32.716 total`

With this patch: `0.34s user 0.20s system 2% cpu 24.880 total`

The Go command prior to this patch hadn't finished after 2 minutes.

https://jira.mesosphere.com/browse/DCOS_OSS-5357

### DIFF
--- a/pkg/mesos/taskio.go
+++ b/pkg/mesos/taskio.go
@@ -287,9 +287,14 @@ func (t *TaskIO) attachContainerInput() error {
 			}()
 
 			for {
-				// TODO(bamarni): investigate on why "test_task:test_attach" fails if the buffer size below
-				// is greater than 1, this might be an issue with the term package or the way we use it.
-				buf := make([]byte, 1)
+				bufLen := 512
+				if t.opts.TTY {
+					// TODO(bamarni): investigate on why "test_task:test_attach" fails if the buffer size below
+					// is greater than 1, this looks like an issue with the term package we're using, which doesn't
+					// work properly when reading more than 1 char at a time.
+					bufLen = 1
+				}
+				buf := make([]byte, bufLen)
 				n, err := t.opts.Stdin.Read(buf)
 				if _, ok := err.(term.EscapeError); ok {
 					t.cancelFunc()
--- a/pkg/mesos/taskio.go
+++ b/pkg/mesos/taskio.go
@@ -287,7 +287,10 @@ func (t *TaskIO) attachContainerInput() error {
 			}()
 
 			for {
+				// Use a buffer with a reasonable size in order to perform well when STDIN is piped and has
+				// a significant size. See https://jira.mesosphere.com/browse/DCOS_OSS-5357
 				bufLen := 512
+
 				if t.opts.TTY {
 					// TODO(bamarni): investigate on why "test_task:test_attach" fails if the buffer size below
 					// is greater than 1, this looks like an issue with the term package we're using, which doesn't


pkg/mesos/taskio.go
```diff
@@ -287,9 +287,14 @@ func (t *TaskIO) attachContainerInput() error {
 			}()
 
 			for {
-				// TODO(bamarni): investigate on why "test_task:test_attach" fails if the buffer size below
-				// is greater than 1, this might be an issue with the term package or the way we use it.
-				buf := make([]byte, 1)
+				bufLen := 512
```
vespian, line 292: Why 512 in particular?

bamarni: It's similar to https://github.com/mesos/mesos-go/blob/3c2ae7f09d530e0ecd62acd2c31a4d73ed9bdf55/api/v1/cmd/msh/app/app.go#L517

For the record the Python CLI had 1024: https://github.com/dcos/dcos-core-cli/blob/1c5f9faffe757abef90ec337fdacaf7b6c01c8f4/python/lib/dcos/dcos/mesos.py#L1644

Do you have any suggestions regarding this?

bamarni: Using the same file input, it took 22, then 23 seconds with the value bumped to 1024. Maybe there is a slight improvement if we bump it, though it also increases memory usage. I think the main improvement is not to leave it to something as low as 1.

bamarni:

pkg/mesos/taskio.go
```diff
@@ -287,9 +287,17 @@ func (t *TaskIO) attachContainerInput() error {
...
 			for {
-				// TODO(bamarni): investigate on why "test_task:test_attach" fails if the buffer size below
-				// is greater than 1, this might be an issue with the term package or the way we use it.
-				buf := make([]byte, 1)
+				// Use a buffer with a reasonable size in order to perform well when STDIN is piped and has
+				// a significant size. See https://jira.mesosphere.com/browse/DCOS_OSS-5357
+				bufLen := 512
+
+				if t.opts.TTY {
+					// TODO(bamarni): investigate on why "test_task:test_attach" fails if the buffer size below
```
jdef, line 295: is there a JIRA for this TODO? if so, mind linking it to this PR?

bamarni: There is none, on the other hand I'm not sure we should have one. As there is no known bug or performance issue because of this, I don't see how it would make it to one of our sprint.

As far as I could tell when testing manually it is reading byte by byte when the terminal is in raw mode anyways. The issue was only seen on this integration test with a pty, which made the CLI read more than 1 byte at a time: https://github.com/dcos/dcos-core-cli/blob/6c5d73f51087aee9527b03464f6b91719ab87100/python/lib/dcoscli/tests/integrations/test_task.py#L437-L438

bamarni:

bamarni:

bamarni:

PR #303, opened 2019-07-23
Fix string properties parsing in jsonitem without schema

bamarni:
Currently, the following command would fail:

$ dcos marathon app update /kit/kat role=slave_public
Error while fetching [https://18.236.223.84/service/marathon/v2/apps/kit/kat]: HTTP 422: "{"message":"Object is not valid","details":[{"path":"/role","errors":["got https://slave_public, expected one of: [slave_public, kit]"]}]}".

It also happens with other string fields like `cmd`. The CLI is adding
a `https://` prefix when it shouldn't.

https://jira.mesosphere.com/browse/DCOS-55320


bamarni:

armandgrillet (approved):

PR #310, opened 2019-07-31
Various improvements on 'dcos quota'.

armandgrillet:


pkg/cmd/quota/quota_create.go
```diff
@@ -17,6 +19,17 @@ func newCmdQuotaCreate(ctx api.Context) *cobra.Command {
 		Use:   "create <group>",
 		Short: "Create a quota",
 		Args:  cobra.ExactArgs(1),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			// We exclude the first character in case it starts with "/" and then check for "/".
+			if strings.Contains(args[0][1:], "/") {
```
bamarni: `dcos quota create ""` would cause a panic. Even though running this command doesn't really make sense, from the UX perspective we shouldn't fail with stack traces.

pkg/cmd/quota/quota_list.go
```diff
@@ -64,9 +65,49 @@ func newCmdQuotaList(ctx api.Context) *cobra.Command {
...
+						}
+						percent := fmt.Sprintf("%.2f%%", cons*100/limit)
+						var desc string
+						switch info {
+						case "cpus", "gpus":
+							desc = fmt.Sprintf("%.0f of %.0f Cores", cons, limit)
+						case "mem", "disk":
+
+							if limit < 1000 {
+								desc = fmt.Sprintf("%.0fMiB of %.0fMiB", cons, limit)
```
bamarni: do we want a space between the value an MiB? That's how it is in the UI.

armandgrillet: Screenshot 2019-07-31 à 17 46 11
I used the Invision website but you're correct, the new UI shows spaces.

bamarni (approved):

PR #311, opened 2019-08-02
Updated 'dcos quota create' to error out less often.

armandgrillet:


pkg/marathon/client.go
```diff
@@ -41,8 +41,9 @@ func NewClient(ctx api.Context) (*Client, error) {
 	return &Client{API: client, baseURL: baseURL}, nil
 }
 
-// GroupsAsQuotas returns the Marathon groups' names as quotas.
-func (c *Client) GroupsAsQuotas() (map[string]bool, error) {
+// GroupsWithoutRootSlash returns the Marathon groups' names as a map without the first "/".
+// This makes it easy to match Marathon groups with Mesos roles that cannot start with "/".
+func (c *Client) GroupsWithoutRootSlash() (map[string]bool, error) {
```
bamarni, line 46: Not blocking but I think this method is too specific to be in a marathon client, I'd leave the slash striping logic outside of the client.

bamarni (approved):

PR #314, opened 2019-08-07
 Don't fail on 500 errors when forwarding STDIN

bamarni:
This happens when the container exits before the call to attach the
container input is done, in such case Mesos will return a 500 error
when the call is done later on, we should fail silently in such cases
and continue to consume the container output.

This was handled in Python here:

https://github.com/dcos/dcos-core-cli/blob/793c47ce90f49a4cdf6b3b47d9f358ede510a4ca/python/lib/dcos/dcos/mesos.py#L1569-L1576

https://jira.mesosphere.com/browse/DCOS_OSS-5432


bamarni:

bamarni:

armandgrillet (approved):

rgo3 (approved):

PR #317, opened 2019-08-12
Add diagnostics command

br-lewis:
This adds a `diagnostics` command along with `diagnostics list` as the first subcommand to use the new Diagnostics API.

https://jira.mesosphere.com/browse/DCOS_OSS-5397
https://jira.mesosphere.com/browse/DCOS_OSS-5306

### DIFF
--- a/completion/bash/diagnostics.sh
+++ b/completion/bash/diagnostics.sh
@@ -1,0 +1,52 @@
+
+_dcos_diagnostics() {
+    local i command
+
+    if ! __dcos_default_command_parse; then
+        return
+    fi
+
+    local flags=(
+    "--help"
+    )
+
+    local commands=(
+    "list"
+    )
+
+    if [ -z "$command" ]; then
+        case "$cur" in
+            --*)
+                __dcos_handle_compreply "${flags[@]}"
+                ;;
+            *)
+                __dcos_handle_compreply "${commands[@]}"
+                ;;
+        esac
+        return
+    fi
+
+    __dcos_handle_subcommand
+}
+
+_dcos_diagnostics_list() {
+    local i command
+
+    if ! __dcos_default_command_parse; then
+        return
+    fi
+
+    local flags=(
+    "--help"
+    "--json"
+    )
+
+    if [ -z "$command" ]; then
+        case "$cur" in
+            --*)
+                __dcos_handle_compreply "${flags[@]}"
+                ;;
+        esac
+        return
+    fi
+}
--- a/pkg/cmd/dcos.go
+++ b/pkg/cmd/dcos.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"github.com/dcos/dcos-cli/api"
+	"github.com/dcos/dcos-core-cli/pkg/cmd/diagnostics"
 	"github.com/dcos/dcos-core-cli/pkg/cmd/job"
 	"github.com/dcos/dcos-core-cli/pkg/cmd/node"
 	"github.com/dcos/dcos-core-cli/pkg/cmd/pkg"
@@ -30,6 +31,7 @@ func NewDCOSCommand(ctx api.Context) *cobra.Command {
 		quota.NewCommand(ctx),
 		service.NewCommand(ctx),
 		task.NewCommand(ctx),
+		diagnostics.NewCommand(ctx),
 	)
 
 	// This follows the CLI design guidelines for help formatting.
--- a/pkg/cmd/diagnostics/diagnostics.go
+++ b/pkg/cmd/diagnostics/diagnostics.go
@@ -1,0 +1,17 @@
+package diagnostics
+
+import (
+	"github.com/dcos/dcos-cli/api"
+	"github.com/spf13/cobra"
+)
+
+// NewCommand creates and returns a diagnostics command with its subcommands
+// already added.
+func NewCommand(ctx api.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "diagnostics",
+		Short: "Create and manage DCOS diagnostics bundles",
+	}
+	cmd.AddCommand(newDiagnosticsListCommand(ctx))
+	return cmd
+}
--- a/pkg/cmd/diagnostics/diagnostics_list.go
+++ b/pkg/cmd/diagnostics/diagnostics_list.go
@@ -1,0 +1,53 @@
+package diagnostics
+
+import (
+	"encoding/json"
+
+	"github.com/dcos/dcos-cli/api"
+	"github.com/dcos/dcos-cli/pkg/cli"
+	diagnostics "github.com/dcos/dcos-core-cli/pkg/diagnostics/v2"
+	"github.com/dcos/dcos-core-cli/pkg/pluginutil"
+	"github.com/dustin/go-humanize"
+	"github.com/spf13/cobra"
+)
+
+func newDiagnosticsListCommand(ctx api.Context) *cobra.Command {
+	var jsonOutput bool
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List available bundles",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			c := pluginutil.HTTPClient("")
+			client := diagnostics.NewClient(c)
+			bundles, err := client.List()
+			if err != nil {
+				return err
+			}
+
+			if jsonOutput {
+				enc := json.NewEncoder(ctx.Out())
+				enc.SetIndent("", "    ")
+				return enc.Encode(bundles)
+			}
+
+			tableHeader := []string{"ID", "STATUS", "CREATED", "SIZE"}
+			table := cli.NewTable(ctx.Out(), tableHeader)
+
+			for _, b := range bundles {
+				size := humanize.Bytes(uint64(b.Size))
+				stopped := humanize.Time(b.Stopped)
+
+				fields := []string{b.ID, b.Status.String(), stopped, size}
+				table.Append(fields)
+			}
+
+			table.Render()
+
+			return nil
+		},
+	}
+
+	cmd.Flags().BoolVar(&jsonOutput, "json", false, "Print in json format")
+
+	return cmd
+}
--- a/pkg/diagnostics/v2/client.go
+++ b/pkg/diagnostics/v2/client.go
@@ -1,0 +1,62 @@
+package v2
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/dcos/dcos-cli/pkg/httpclient"
+)
+
+// Bundle represents a bundle object received from the diagnostics API
+type Bundle struct {
+	ID      string    `json:"id,omitempty"`
+	Size    int64     `json:"size,omitempty"` // length in bytes for regular files; 0 when Canceled or Deleted
+	Status  Status    `json:"status"`
+	Started time.Time `json:"started_at,omitempty"`
+	Stopped time.Time `json:"stopped_at,omitempty"`
+	Errors  []string  `json:"errors,omitempty"`
+}
+
+// Client is a REST API wrapper around the new Diagnostics API.
+type Client struct {
+	http *httpclient.Client
+}
+
+// NewClient creates a new Client.
+func NewClient(baseClient *httpclient.Client) *Client {
+	return &Client{
+		http: baseClient,
+	}
+}
+
+// List gets a list of all cluster bundles.
+func (c *Client) List() ([]Bundle, error) {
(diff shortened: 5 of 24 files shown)


janisz (approved):

pkg/cmd/dcos.go
```diff
@@ -30,6 +31,7 @@ func NewDCOSCommand(ctx api.Context) *cobra.Command {
 		quota.NewCommand(ctx),
 		service.NewCommand(ctx),
 		task.NewCommand(ctx),
+		diagnostics.NewDiagnosticsCommand(ctx),
```
armandgrillet: Should be `NewCommand(ctx)` to go with the others.

pkg/cmd/diagnostics/diagnostics_list.go
```diff
@@ -0,0 +1,60 @@
...
+	table.Render()
+
+	return nil
+}
+
+func newDiagnosticsListCommand(ctx api.Context) *cobra.Command {
+	var jsonOutput bool
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "Lists available bundles",
```
armandgrillet: For coherence with the other commands, use `List available bundles`

pkg/cmd/diagnostics/diagnostics_list.go
```diff
@@ -0,0 +1,60 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			c := pluginutil.HTTPClient("")
+			return listDiagnosticsBundles(ctx, c, jsonOutput)
```
armandgrillet: As this is only called once, we can have the code of `listDiagnosticsBundles` in `RunE` directly.

pkg/diagnostics/v2/client.go
```diff
@@ -0,0 +1,43 @@
...
+
+// List gets a list of all cluster bundles.
+func (c *Client) List() ([]rest.Bundle, error) {
+	resp, err := c.http.Get("/system/health/v1/diagnostics")
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	switch {
```
armandgrillet: Can we go for our standard pattern or does diagnostics return something else than a `200` in some cases?
```
switch resp.StatusCode {
case 200:
    ...
default:
    return nil, fmt.Errorf("unexpected status code %d", resp.StatusCode)
}

File: scripts/plugin/package_plugin.py [hunk 5]
```diff
@@ -45,6 +45,11 @@
 name = "task"
 path = "bin/dcos{0}"
 description = "Manage DC/OS tasks"
+
+[[commands]]
```
armandgrillet: This shouldn't be here yet, let's wait for the subcommand to be complete.

armandgrillet (changes requested):

pkg/diagnostics/v2/client.go
```diff
@@ -0,0 +1,43 @@
+package v2
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/dcos/dcos-cli/pkg/httpclient"
+	"github.com/dcos/dcos-diagnostics/api/rest"
```
makkes: I'm not sure we should add this strict coupling here but rather only depend on HTTP and JSON as an API.

pkg/diagnostics/v2/client_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/dcos/dcos-core-cli/pkg/pluginutil"
+	"github.com/dcos/dcos-diagnostics/api/rest"
```
makkes: I'm not sure we should add this strict coupling here but rather only depend on HTTP and JSON as an API.

makkes (changes requested):

pkg/diagnostics/v2/client.go
```diff
@@ -0,0 +1,43 @@
+package v2
```
rgo3, line 1: Why is this package called `v2`? The API thats called still seems to be `v1`. Is there a chance of merging this file with the already existing client.go and therefore also merge the two `List()` methods? Or are they dealing with different types of Bundles?

br-lewis: It's using a new API at a slightly different endpoint which returns different types. It's pretty much entirely unrelated to the old one aside from them both talking to the same service to accomplish similar things.

rgo3:
A few general thoughts:
- I think this PR fails to build due to the import of `dcos-diagnostics`. AFAIK we have seen this failure in the past and decided to just "do a little copying" of types needed for the responses of `dcos-diagnostics`. We also preferred that over importing because `dcos-diagnostics` is a huge dependency. @ArmandGrillet probably can confirm this.
- Thanks for adding the unit tests, can we add an TODO of adding some integration tests too once this command becomes user-facing?

pkg/cmd/diagnostics/diagnostics_list.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+func newDiagnosticsListCommand(ctx api.Context) *cobra.Command {
+	var jsonOutput bool
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List available bundles",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			c := pluginutil.HTTPClient("")
+			client := diagnostics.NewClient(c)
```
makkes, line 21:
```suggestion
			client := v2.NewClient(c)
```
I wonder why the checks didn't fail for this PR...

makkes (changes requested):

pkg/cmd/diagnostics/diagnostics_list.go
rgo3: Why should it fail? `diagnostics` referred to the `v2` package in the import, which is ok I think.

completion/bash/diagnostics.sh
```diff
@@ -0,0 +1,52 @@
+
+_dcos_diagnostics() {
+	local i command
```
bamarni: There is a mix of tabs and spaces in the file

pkg/cmd/diagnostics/diagnostics.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/dcos/dcos-cli/api"
+)
+
+// NewCommand creates and returns a diagnostics command with its subcommands
+// already added.
+func NewCommand(ctx api.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:    "diagnostics",
+		Hidden: true,
```
bamarni:
```suggestion
```

This is not really necessary, we could add this to any command and it would still be shown in the help menu. The plugin.toml is authoritative here.

pkg/cmd/diagnostics/diagnostics_list.go
```diff
@@ -0,0 +1,55 @@
...
+	"encoding/json"
+
+	"github.com/spf13/cobra"
+
+	"github.com/dcos/dcos-cli/api"
+	"github.com/dcos/dcos-cli/pkg/cli"
+	"github.com/dustin/go-humanize"
+
+	diagnostics "github.com/dcos/dcos-core-cli/pkg/diagnostics/v2"
+	"github.com/dcos/dcos-core-cli/pkg/pluginutil"
```
bamarni, line 9: Do you use a specific rule for splitting imports here? Currently we only use 2 groups (one for stdlib and another one for the rest).

pkg/cmd/diagnostics/diagnostics_list.go
```diff
@@ -0,0 +1,55 @@
...
+				return err
+			}
+
+			if jsonOutput {
+				enc := json.NewEncoder(ctx.Out())
+				enc.SetIndent("", "    ")
+				return enc.Encode(bundles)
+			}
+
+			tableHeader := []string{"ID", "STATUS", "STOPPED", "SIZE"}
```
bamarni:
```suggestion
			tableHeader := []string{"ID", "STATUS", "CREATED", "SIZE"}
```

Even though the component has the concept of start and end dates, on a list I think it is more clear to call it `CREATED` or `DATE`. If a user wants to know more details about how long it took etc. they can use the JSON output.

pkg/diagnostics/v2/client.go
```diff
@@ -0,0 +1,53 @@
...
+	switch resp.StatusCode {
+	case http.StatusOK:
+		var bundles []Bundle
+		err = json.NewDecoder(resp.Body).Decode(&bundles)
+		if err != nil {
+			return nil, err
+		}
+		return bundles, err
+	default:
+		return nil, fmt.Errorf("received unexpected status code %d", resp.StatusCode)
```
bamarni: Is there a list of known HTTP error codes that the diagnostics API might return and can be tied to more informative error messages?

At the very least, we should handle 401 and 403 status codes, see https://github.com/dcos/dcos-core-cli/blob/1.14-patch.x/pkg/diagnostics/client.go for example, with the httpResponseToError helper.

The generic 401 and 403 messages are handled here : https://github.com/dcos/dcos-core-cli/blob/1.14-patch.x/cmd/dcos/main.go

pkg/cmd/diagnostics/diagnostics_list.go
makkes: Ah, didn't see that import. All good. I was never really here.

makkes (approved):

armandgrillet (approved):

bamarni:
Might be too late for this first version but we should consider adding the spec to https://github.com/dcos/client-go at some point and using the generated client, as we're trying to adopt it more across our projects.

pkg/diagnostics/v2/status.go
```diff
@@ -0,0 +1,64 @@
...
+// Status represents an bundle status
+type Status int
+
+// Possible statuses
+const (
+	Unknown    Status = iota // No information about this bundle
+	Started                  // Diagnostics is preparing
+	InProgress               // Diagnostics in progress
+	Done                     // Diagnostics finished and the file is ready to be downloaded
+	Canceled                 // Diagnostics has been cancelled
```
drewvanstone: s/cancelled/canceled/g

makkes: https://www.grammarly.com/blog/canceled-vs-cancelled/ says both are correct. :-D 🇬🇧🇺🇸

pkg/cmd/diagnostics/diagnostics_list.go
br-lewis: I do though it looks like I broke it here. Normally I'll do 3 groups: stdlib, external imports, and imports within this project. I don't have any real reason for that besides it looking nicer to me so if you want it down to 2 groups, that's fine as well.

pkg/diagnostics/v2/client.go
br-lewis: In other endpoints yes but not for `List`, I've added the use of the `httpResponseToError` helper though.

pkg/cmd/diagnostics/diagnostics_list.go
bamarni: In the future we could enforce a rule with goimports and have a CI check, in the meantime let's stick to 2 groups for consistency.

bamarni (approved):
LGTM

bamarni:

makkes (changes requested):
@br-lewis can you please change the commit message to sth. reasonable :-)

br-lewis:

armandgrillet:
Please bump your PR to fix CI.

bamarni:

bamarni:
Re-triggering should fix the build. Seems like we got a race condition in our integration tests...

We have a 2 step process where we build the Go binary in the `mesos-ubuntu` node, and then run integration tests in the `mesos` node.

Each node will checkout the codebase and rebase against the target branch to make sure it picks the latest changes. However here, we built the Go binary on top of 1896b885581b8db2553f8e6f66a783c9600d505e (as highlighted in the build log with `Commit message: "Merge a2a21c9f7a05e7b7a25226432264d18bfc142e08 into 1896b885581b8db2553f8e6f66a783c9600d505e"`).

But later on the integration tests were ran on top of cc31576da855bf651ea3710b476ed0c660e7eb6c (which was merged in the meantime, and as highlighted with the build log message `Commit message: "Merge a2a21c9f7a05e7b7a25226432264d18bfc142e08 into cc31576da855bf651ea3710b476ed0c660e7eb6c").`

So integration tests were expecting the new Go `dcos package search` command but it was actually using the Python one, thus the `AssertionError: assert b'no packages found' in b'No packages found.\n'` error.

PR #318, opened 2019-08-12
taskIO fixes

bamarni:
We've been using a 1 byte buffer with the proxy reader to get around a
failure in one of our integration tests.

However this creates an error condition, when the CLI matches a partial
escape sequence, it cannot recover from it and will fail with a stack
trace.

This can be reproduced by typing eg. `ctrl-p+A`, when the proxy reader
reads `ctrl-p`, it will not return any data as it's matching the
beginning of the `ctrl-p+ctrl-q` escape sequence. However, the next
character `A` should lead to the reader returning `ctrl-p` and then `A`,
either on the current read if the buffer has enough capacity, or on the
next read. However this doesn't seem to work.

Instead of using a 1 byte buffer, I've patched the escape proxy reader
to work with a base reader potentially returning more than 1 byte (eg. a
pty), and will submit a PR to Docker.

https://jira.mesosphere.com/browse/DCOS_OSS-5443


bamarni:

bamarni:
For reference the PR is there: https://github.com/moby/moby/pull/39728

For now I've moved the updated escape proxy to our codebase.

bamarni:

pkg/mesos/taskio.go
```diff
@@ -298,30 +297,19 @@ func (t *TaskIO) attachContainerInput() error {
...
-					// work properly when reading more than 1 char at a time.
-					bufLen = 1
-				}
-				buf := make([]byte, bufLen)
+				buf := make([]byte, 512)
 				n, err := t.opts.Stdin.Read(buf)
-				if _, ok := err.(term.EscapeError); ok {
-					t.cancelFunc()
-					t.exitSequenceDetected = true
-					return
```
bamarni, line 314: This change is unrelated, there were a race condition between sending an empty input data to indicate EOF and calling cancelFunc(), which can cause the input connection to be closed before Mesos received the EOF message.

Moving cancelFunc() to after the input HTTP call finished (once the acif channel is closed) makes sure Mesos has received EOF on STDIN and we only cancel the call for container output.

armandgrillet (approved):

PR #320, opened 2019-08-14
Add diagnostics download command

br-lewis:
Addresses https://jira.mesosphere.com/browse/DCOS_OSS-5307


pkg/cmd/diagnostics/diagnostics_download.go
```diff
@@ -0,0 +1,51 @@
...
+		Short: "Download diagnostics bundle",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			c := pluginutil.HTTPClient("")
+			client := diagnostics.NewClient(c)
+
+			id := args[0]
+
+			if outputPath == "" {
+				outputPath = fmt.Sprintf("%s.zip", id)
```
makkes, line 28: I think we should sanitize this user input in some way to prevent malicious behaviour.

br-lewis: We could limit `id` to contain only alphanumerics and dashes but I'm not sure what we could do to `outputPath` to prevent issues. Maybe not allow it to have `..` but if we're worried about malicious use, an attacker could just as easily `cd` to their target directory first (or `/`).

Maybe I'm not seeing the potential malicious behavior you are.

pkg/diagnostics/v2/client.go
```diff
@@ -52,6 +53,34 @@ func (c *Client) List() ([]Bundle, error) {
...
+	case http.StatusOK:
+		_, err := io.Copy(dst, resp.Body)
+		if err != nil {
+			return err
+		}
+		return nil
+	case http.StatusNotFound:
+		return fmt.Errorf("no bundle %s found", id)
+	case http.StatusNotModified:
+		return fmt.Errorf("bundle %s canceled or not completed", id)
```
janisz: We do not send Not Modified for bundle download only for Delete

pkg/diagnostics/v2/client.go
```diff
@@ -52,6 +53,32 @@ func (c *Client) List() ([]Bundle, error) {
...
+		return err
+	}
+	defer resp.Body.Close()
+
+	switch resp.StatusCode {
+	case http.StatusOK:
+		_, err := io.Copy(dst, resp.Body)
+		if err != nil {
+			return err
+		}
```
bamarni: the if block can be removed to just have `return err`

pkg/cmd/diagnostics/diagnostics_download.go
```diff
@@ -0,0 +1,51 @@
...
+			outFile, err := os.Create(outputPath)
+			if err != nil {
+				return err
+			}
+			defer outFile.Close()
+
+			err = client.Download(id, outFile)
+			if err != nil {
+				outFile.Close()
+				os.Remove(outputPath)
```
bamarni: I don't think cleanup is needed here: 
- user can still call the command again and it will overwrite outputPath if it already exists
- having a partially downloaded or corrupted file leaves room for debugging stuff in case of errors, also, given that diagnostics files can be quite big I can imagine that in the future we might be able to resume a broken download, like eg. `wget --continue`
- it keeps the command code simple

pkg/cmd/diagnostics/diagnostics_download.go
bamarni: It looks fine as is IMO, users should be able to set any path as long as they have permission to create a file there.

pkg/diagnostics/v2/client_test.go
```diff
@@ -52,3 +54,79 @@ func TestEmptyList(t *testing.T) {
 
 	assert.Empty(t, bundles)
 }
+
+func TestDownload(t *testing.T) {
```
bamarni, line 58: :100:

pkg/cmd/diagnostics/diagnostics_download.go
bamarni: ~However I wouldn't add the `.zip` suffix automatically. If I pass an explicit flag I wouldn't expect this behavior. That's similar to what we do in other commands, eg. `dcos license audit get -o /path/to/file`~

I misread actually, this is the default value not when the flag is passed.

pkg/cmd/diagnostics/diagnostics_download.go
```diff
@@ -0,0 +1,48 @@
...
+				outputPath = fmt.Sprintf("%s.zip", id)
+			}
+
+			outFile, err := os.Create(outputPath)
+			if err != nil {
+				return err
+			}
+			defer outFile.Close()
+
+			err = client.Download(id, outFile)
```
bamarni: `return client.Download(id, outFile)`

br-lewis:

pkg/cmd/diagnostics/diagnostics_download.go
makkes: @br-lewis, you're probably right. In the end the file will be created on the user's filesystem, so he could potentially only harm himself.

makkes (approved):

PR #323, opened 2019-08-15
Change created column to use started instead of stopped

br-lewis:
https://jira.mesosphere.com/browse/DCOS_OSS-5455


pkg/cmd/diagnostics/diagnostics_list.go
```diff
@@ -35,9 +35,9 @@ func newDiagnosticsListCommand(ctx api.Context) *cobra.Command {
 
 			for _, b := range bundles {
 				size := humanize.Bytes(uint64(b.Size))
-				stopped := humanize.Time(b.Stopped)
+				created := humanize.Time(b.Started)
```
armandgrillet, line 38: Should we use `STARTED` in the table as the JSON key is `started_at`?

armandgrillet (approved):

bamarni (approved):

PR #326, opened 2019-08-19
DCOS_OSS-5305 Use new API for creating bundles

makkes:


pkg/diagnostics/v2/client.go
```diff
@@ -76,6 +79,36 @@ func (c *Client) Download(id string, dst io.Writer) error {
 	}
 }
 
+// Create creates a new cluster bundle and returns its ID.
+func (c *Client) Create() (string, error) {
+	newID := uuid.NewV4()
+	req, err := c.http.NewRequest("PUT", fmt.Sprintf("%s/%s", baseURL, newID.String()), nil)
```
bamarni:
```suggestion
	req, err := c.http.NewRequest("PUT", fmt.Sprintf("%s/%s", baseURL, uuid.NewV4()), nil)
```

Just curious, why is the client expected to generate the ID, instead of having a POST endpoint which would create the id and return it?

pkg/diagnostics/v2/client_test.go
```diff
@@ -55,6 +59,43 @@ func TestEmptyList(t *testing.T) {
...
 }
 
+func TestCreateHappyPath(t *testing.T) {
+	var newID string
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		re := regexp.MustCompile("^/system/health/v1/diagnostics/([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})$")
+		urlMatch := re.FindStringSubmatch(r.URL.Path)
+		newID = urlMatch[1]
+		require.True(t, len(urlMatch) == 2, fmt.Sprintf("URL %s doesn't match %s", r.URL.Path, re.String()))
+		require.Equal(t, "PUT", r.Method)
```
bamarni: We should use assert here, as require calls [FailNow](https://golang.org/pkg/testing/#T.FailNow), which should only be called from the main goroutine running the test.

pkg/diagnostics/v2/client_test.go
```diff
@@ -55,6 +59,43 @@ func TestEmptyList(t *testing.T) {
 	assert.Empty(t, bundles)
 }
 
+func TestCreateHappyPath(t *testing.T) {
+	var newID string
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		re := regexp.MustCompile("^/system/health/v1/diagnostics/([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})$")
+		urlMatch := re.FindStringSubmatch(r.URL.Path)
+		newID = urlMatch[1]
+		require.True(t, len(urlMatch) == 2, fmt.Sprintf("URL %s doesn't match %s", r.URL.Path, re.String()))
```
bamarni: Maybe simpler to rely on https://godoc.org/github.com/satori/go.uuid#FromString for validating the uuid.

go.mod
```diff
@@ -23,6 +23,7 @@ require (
 	github.com/satori/go.uuid v1.2.0
 	github.com/sirupsen/logrus v1.4.2
 	github.com/spf13/cobra v0.0.4
+	github.com/spf13/pflag v1.0.3
```
janisz, line 26: Where do we need this?

janisz (approved):

pkg/diagnostics/v2/client.go
makkes: Great question. It's a design decision I'm not particularly happy with but haven't been involved too much. Probably @janisz might be able to elaborate on it.

go.mod
makkes: `pkg/pluginutil/pluginutil.go`

janisz (approved):

pkg/diagnostics/v2/client.go
```diff
@@ -76,6 +79,35 @@ func (c *Client) Download(id string, dst io.Writer) error {
 	}
 }
 
+// Create creates a new cluster bundle and returns its ID.
+func (c *Client) Create() (string, error) {
+	req, err := c.http.NewRequest("PUT", fmt.Sprintf("%s/%s", baseURL, uuid.NewV4().String()), nil)
```
bamarni, line 84: nit: `.String()` shouldn't be necessary here

bamarni (approved):

pkg/diagnostics/v2/client.go
makkes: Yes sure. I like explicit code, though, but I have no clue whether there's a beat practice for this case.

PR #329, opened 2019-08-19
DCOS_OSS-5430 Use new API for deleting bundle

makkes:


br-lewis (approved):

pkg/diagnostics/v2/client.go
```diff
@@ -108,6 +108,30 @@ func (c *Client) Create() (string, error) {
 	}
 }
 
+// Delete deletes the given cluster bundle.
+func (c *Client) Delete(id string) error {
+	req, err := c.http.NewRequest("DELETE", fmt.Sprintf("%s/%s", baseURL, id), nil)
```
janisz, line 113: How about extract `fmt.Sprintf("%s/%s", baseURL, id)` into a function?
```suggestion
	req, err := c.http.NewRequest("DELETE", bundleUrl(id), nil)
```

pkg/diagnostics/v2/client.go
```diff
@@ -108,6 +108,30 @@ func (c *Client) Create() (string, error) {
...
+	if err != nil {
+		return err
+	}
+	resp, err := c.http.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	switch resp.StatusCode {
```
janisz, line 123: We can extract this function into a function https://github.com/dcos/dcos-diagnostics/pull/131#pullrequestreview-265278839

janisz (changes requested):
I have mixed feeling about this. We are just doing the same work we did in https://github.com/dcos/dcos-diagnostics/blob/master/api/rest/client.go even with the exact same comments. Why just not simply copy that code if we do not want to introduce a dependecny to dcos-diagnostics in cli?

pkg/diagnostics/v2/client.go
makkes: I'm not convinced we should introduce a level of abstraction here for sth. trivial like this.

pkg/diagnostics/v2/client.go
makkes: We're handling the different status codes differently in each method:

* In `Delete` (here): 404 and 304 are handled explicitly and on 200 `nil` is returned.
* In `Create`: On 200 the method decodes the body and returns the ID, all other codes are generic/unknown errors.
* In `Download`: On 200 the body is written to the `dst` parameter, 404 and 500 are handled explicitly.

The extracted function would have to be so generic that IMHO it's not worth it having it in the first place.

janisz (approved):

PR #330, opened 2019-08-20
Added Go 'dcos package repo' subcommands.

armandgrillet:


pkg/cmd/pkg/package_repo_add.go
```diff
@@ -13,7 +15,16 @@ func newCmdPackageRepoAdd(ctx api.Context) *cobra.Command {
...
+			}
+
+			_, err = c.PackageAddRepo(args[0], args[1], index)
+			if err != nil {
+				return err
+			}
+			return nil
 		},
 	}
 	cmd.Flags().IntVar(&index, "index", 0, "Numerical position in the package repository list. Package repositories are searched in descending order.")
```
bamarni, line 30: Normally the default here is not to send the index, then Cosmos will add it at the last position. Passing `--index=0` should be done explicitly.

This means we need to make the difference between the zero value and a missing value in client-go.

pkg/cmd/pkg/package_repo_import.go
```diff
@@ -10,8 +18,49 @@ func newCmdPackageRepoImport(ctx api.Context) *cobra.Command {
...
+
+			for index, repo := range repo.Repositories {
+				if repo.Name == "" || repo.Uri == "" {
+					fmt.Fprintf(ctx.Out(), "Repo missing name or uri. Skipping.")
+					continue
+				}
+
+				_, err = c.PackageAddRepo(repo.Name, repo.Uri, index)
+				if err != nil {
+					fmt.Fprintf(ctx.ErrOut(), "Error (%s) while adding repo '%s' (%s). Skipping.\n", err, repo.Name, repo.Uri)
```
bamarni, line 57: `ctx.Logger().Errorf()`, does the python CLI behaves like this? Succeeding even though some repos were not added.

pkg/cmd/pkg/package_repo_import.go
```diff
@@ -10,8 +18,49 @@ func newCmdPackageRepoImport(ctx api.Context) *cobra.Command {
...
+			}
+
+			c, err := cosmos.NewClient(ctx, pluginutil.HTTPClient(""))
+			if err != nil {
+				return err
+			}
+
+			for index, repo := range repo.Repositories {
+				if repo.Name == "" || repo.Uri == "" {
+					fmt.Fprintf(ctx.Out(), "Repo missing name or uri. Skipping.")
```
bamarni, line 51: Should go to stderr, maybe use the logger and warning level

pkg/cmd/pkg/package_repo_import.go
```diff
@@ -10,8 +18,49 @@ func newCmdPackageRepoImport(ctx api.Context) *cobra.Command {
...
+					continue
+				}
+
+				_, err = c.PackageAddRepo(repo.Name, repo.Uri, index)
+				if err != nil {
+					fmt.Fprintf(ctx.ErrOut(), "Error (%s) while adding repo '%s' (%s). Skipping.\n", err, repo.Name, repo.Uri)
+					continue
+				}
+				fmt.Fprintf(ctx.Out(), "Added repo '%s' (%s) at index %d", repo.Name, repo.Uri, index)
```
bamarni, line 60: As an informative message, this should go to stderr (see https://github.com/dcos/dcos-cli/blob/master/design/style.md#stdoutstderr).

pkg/cmd/pkg/package_repo_import.go
armandgrillet: Yes: https://github.com/dcos/dcos-core-cli/blob/452fb550827118a2cc6b7d75daf6d5cfc3048935/python/lib/dcoscli/dcoscli/package/main.py#L203

pkg/cmd/pkg/package_repo_add.go
ArmandGrillet: It is differentiated in client-go by having 
```
type CosmosPackageAddRepoV1Request struct {
	Name  string `json:"name"`
	Uri   string `json:"uri"`
	Index int32  `json:"index,omitempty"`
}
```
Will differentiate it when calling the command.

bamarni:
https://github.com/dcos/dcos-core-cli/pull/346

PR #331, opened 2019-08-21
Support custom escape sequences

bamarni:
This can be done through the `DCOS_TASK_ESCAPE_SEQUENCE` env var.

The value follows the same semantics as Docker, see:

https://docs.docker.com/engine/reference/commandline/attach/#override-the-detach-sequence

As a follow-up step in dcos/dcos-cli, we will introduce a new
`task.escape_sequence` config option. The plugin invocation system
will also be updated to pass configuration keys for subcommands as
env variables.

https://jira.mesosphere.com/browse/DCOS-57542

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # CHANGELOG
 
+## 1.14-patch.5
+
+* Features
+
+  * Support custom escape sequences for `dcos task attach`
+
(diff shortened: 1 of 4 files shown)


armandgrillet (approved):

armandgrillet:

rgo3 (approved):
LGTM

PR #338, opened 2019-08-23
DCOS_OSS-5438 Add deprecation notice for 'node diagnostics'

makkes:


janisz (approved):

pkg/cmd/node/node_diagnostics.go
```diff
@@ -14,6 +17,12 @@ func newCmdNodeDiagnostics(ctx api.Context) *cobra.Command {
 		Use:   "diagnostics",
 		Short: "Use diagnostics bundles",
 		RunE: func(cmd *cobra.Command, args []string) error {
+
+			_, ok := ctx.EnvLookup(cli.EnvStrictDeprecations)
+			if ok {
+				return fmt.Errorf("unknown command diagnostics")
+			}
+			ctx.Deprecated("This command is deprecated since DC/OS 1.14, please use 'dcos diagnostics' instead.")
```
bamarni: The env check is done in the Deprecated function, it can be removed to just have:

```suggestion
			err := ctx.Deprecated("This command is deprecated since DC/OS 1.14, please use 'dcos diagnostics' instead.")
			if err != nil {
				return err
			}
```

makkes: Then https://github.com/dcos/dcos-core-cli/blob/1.14-patch.x/pkg/cmd/service/service.go#L18 needs a fix for this b/c I looked it up there. :-)

pkg/cmd/node/node_diagnostics_status.go
```diff
@@ -16,6 +16,11 @@ func newCmdNodeDiagnosticsStatus(ctx api.Context) *cobra.Command {
 		Short: "Print diagnostics job status",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			err := ctx.Deprecated("This command is deprecated since DC/OS 1.14, please use 'dcos diagnostics' instead.")
```
drewvanstone: This should be `dcos diagnostics status`, right?

drewvanstone (approved):

pkg/cmd/node/node_diagnostics_status.go
makkes: That doesn't exist. :-D So we just defer the user to the next higher level of the cmd hierarchy.

br-lewis: We could make it or we could point them to `diagnostics list` as well.

br-lewis:
Doesn't deprecation usually not prevent you from using the command? It seems like if we detect it's deprecated the CLI bails immediately.

bamarni:
@br-lewis : it defaults to printing a warning, only when `DCOS_CLI_STRICT_DEPRECATIONS=1` is exported it will fail with an error. 

We introduced this env var to make it easier to detect deprecated usage in scripts, where deprecation messages can easily be missed in long outputs.

pkg/cmd/node/node_diagnostics.go
bamarni: The use-case is different there, as it's not deprecating the command altogether but changing its behavior (printing the help menu instead of showing a list).

PR #342, opened 2019-08-26
Add wait command to diagnostics

br-lewis:
This add a wait command that will wait for a diagnostics bundle to finish.
I also added a spinner while the process waits but this is a little un-unixy so maybe we'll want to remove that before merging the PR.

Addresses: https://jira.mesosphere.com/browse/DCOS_OSS-5488

### DIFF
--- a/completion/bash/diagnostics.sh
+++ b/completion/bash/diagnostics.sh
@@ -15,6 +15,7 @@ _dcos_diagnostics() {
     "delete"
     "download"
     "list"
+    "wait"
     )
 
     if [ -z "$command" ]; then
@@ -125,6 +126,27 @@ _dcos_diagnostics_list() {
         esac
         return
     fi
+}
+
+_dcos_diagnostics_wait() {
+    local i command
+
+    if ! __dcos_default_command_parse; then
+        return
+    fi
+
+    local flags=(
+    "--help"
+    )
+
+    case "$cur" in
+        --*)
+            __dcos_handle_compreply "${flags[@]}"
+            ;;
+        *)
+            __dcos_complete_bundle_ids
+            ;;
+    esac
 }
 
 __dcos_complete_bundle_ids() {
--- a/pkg/cmd/diagnostics/diagnostics_wait.go
+++ b/pkg/cmd/diagnostics/diagnostics_wait.go
@@ -1,0 +1,85 @@
+package diagnostics
+
+import (
+	"errors"
+	"time"
+
+	"github.com/dcos/dcos-cli/api"
+	diagnostics "github.com/dcos/dcos-core-cli/pkg/diagnostics/v2"
+	"github.com/dcos/dcos-core-cli/pkg/pluginutil"
+	"github.com/spf13/cobra"
+)
+
+const pollInterval time.Duration = time.Second
+
+func newDiagnosticsWaitCommand(ctx api.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "wait <bundle-id>",
+		Short: "Wait until the given bundle is completed",
+		Args:  cobra.MaximumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			c := pluginutil.HTTPClient("")
+			client := diagnostics.NewClient(c)
+
+			// this seemed to be the easiest way to manage using two different methods
+			// of getting the right bundle depending on the arguments
+			bundle, err := func() (*diagnostics.Bundle, error) {
+				if len(args) == 0 {
+					return latestBundle(client)
+				}
+				return client.Get(args[0])
+			}()
+			if err != nil {
+				return err
+			}
+
+			if len(args) == 0 {
+				ctx.Logger().Infof("Waiting for bundle %s", bundle.ID)
+			}
+
+			for {
+				if bundle.IsFinished() {
+					break
+				}
+
+				time.Sleep(pollInterval)
+
+				bundle, err = client.Get(bundle.ID)
+				if err != nil {
+					return err
+				}
+			}
+
+			return nil
+		},
+	}
+
+	return cmd
+}
+
+func latestBundle(client *diagnostics.Client) (*diagnostics.Bundle, error) {
+	bundles, err := client.List()
+	if err != nil {
+		return nil, err
+	}
+	if len(bundles) == 0 {
+		return nil, errors.New("no bundles found")
+	}
+
+	// default time.Time is as far back we'll need to worry about anyway so serves as a good starting min
+	var max time.Time
+	var bundle diagnostics.Bundle
+
+	for _, b := range bundles {
+		if b.Type == diagnostics.Cluster && b.Started.After(max) {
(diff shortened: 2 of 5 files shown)


completion/bash/diagnostics.sh
```diff
@@ -15,6 +15,7 @@ _dcos_diagnostics() {
     "delete"
     "download"
     "list"
+	"wait"
```
janisz: Wrong indent

bamarni (commented):
Regarding the command output, I'd leave it empty while waiting, similar to the UNIX `wait` command. If we're concerned in terms of UX, we could have a message like `This command may take a few minutes` in stderr.

It might also be useful to print the bundle ID to stdout, in case the command was launched without an explicit ID.

br-lewis:
I'd prefer to keep it since I think it adds a bit of responsiveness that you wouldn't otherwise get, the `--quiet` flag exists if you don't want to see it and I could modify that to print the bundle ID (instead of nothing, currently).

It's already a bit of an odd command anyway in terms of unix conventions. I think the more "proper" way to do this under unix conventions would be to add a `diagnostics status <bundle>` command then set up a few commands to periodically query the bundle and pipe it in `jq` to check the status until it's finished.

pkg/cmd/diagnostics/diagnostics_wait.go
```diff
@@ -0,0 +1,89 @@
...
+			// only print the ID after verifying that the bundle actually exists
+			fmt.Fprintln(ctx.Out(), bundle.ID)
+
+			// no need to poll if the bundle is done already
+			if bundle.IsFinished() {
+				return nil
+			}
+
+			for {
+				bundle, err = client.Get(bundle.ID)
```
bamarni: If this is moved at the end of the loop, the block above with `// no need to poll if the bundle is done already` can be removed.

Alternatively, it is also fine to keep the command simpler and not optimize here, eg. to only retrieve a bundleID at the beginning of the command (either through the arg or by fetching the latest bundle) and start polling from the loop.

pkg/cmd/diagnostics/diagnostics_wait.go
```diff
@@ -0,0 +1,89 @@
...
+					return latestBundle(client)
+				}
+				return client.Get(args[0])
+			}()
+			if err != nil {
+				return err
+			}
+
+			// only print the ID after verifying that the bundle actually exists
+			fmt.Fprintln(ctx.Out(), bundle.ID)
```
bamarni: Can we move this at the end? It would be easier to see when the command is finished if it prints right before exiting.

pkg/diagnostics/v2/client.go
```diff
@@ -109,6 +114,35 @@ func (c *Client) Create() (string, error) {
...
+		var bundle Bundle
+		err = json.NewDecoder(resp.Body).Decode(&bundle)
+		if err != nil {
+			return nil, err
+		}
+		return &bundle, nil
+	case http.StatusNotFound:
+		return nil, fmt.Errorf("no bundle %s found", id)
+	case http.StatusNotModified:
+		return nil, fmt.Errorf("bundle %s has already been deleted", id)
```
bamarni, line 140: Not directly related to this PR, but is it the right status code for this? Maybe 410 Gone would be more suited.

pkg/cmd/diagnostics/diagnostics_wait.go
br-lewis: I would rather have it print before so that the user can see if it's actually looking at the bundle they expect. Alternatively, I'd say have it print nothing at all. I don't see much use in telling the user what bundle they were waiting on after they spent multiple minutes waiting for it.

pkg/cmd/diagnostics/diagnostics_wait.go
br-lewis: I'm not sure I understand your first suggestion.

pkg/cmd/diagnostics/diagnostics_wait.go
bamarni: Maybe log `Waiting for bundle {id}` with the info level then? Also, how about doing it only when the ID is not explicitly passed?

pkg/cmd/diagnostics/diagnostics_wait.go
bamarni: For the first, I meant:
``` go
for {
	if bundle.IsFinished() {
		break
	}

	time.Sleep(pollInterval)
	
	bundle, err = client.Get(bundle.ID)
	if err != nil {
		return err
	}			
}
```

completion/bash/diagnostics.sh
```diff
@@ -127,6 +128,33 @@ _dcos_diagnostics_list() {
...
+_dcos_diagnostics_wait() {
+    local i command
+
+    if ! __dcos_default_command_parse; then
+        return
+    fi
+
+    local flags=(
+    "--help"
+    "--quiet"
```
bamarni:
```suggestion
```

completion/bash/diagnostics.sh
```diff
@@ -127,6 +128,33 @@ _dcos_diagnostics_list() {
...
+    )
+
+    case "$cur" in
+        --*)
+            __dcos_handle_compreply "${flags[@]}"
+            ;;
+        *)
+            if [ "$prev" == "--output" ]; then
```
bamarni: Should this be removed? I don't see this cli option

bamarni (commented):
left comments on completion script otherwise lgtm

completion/bash/diagnostics.sh
br-lewis: You're right. I copied this from download and forgot to remove this stuff

janisz (approved):

pkg/cmd/diagnostics/diagnostics_wait.go
```diff
@@ -0,0 +1,85 @@
...
+func newDiagnosticsWaitCommand(ctx api.Context) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "wait <bundle-id>",
+		Short: "Wait until the given bundle is completed",
+		Args:  cobra.MaximumNArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			c := pluginutil.HTTPClient("")
+			client := diagnostics.NewClient(c)
+
+			// this seemed to be the easist way to manage using two different methods
```
armandgrillet: s/easist/easiest

pkg/diagnostics/v2/client.go
```diff
@@ -24,6 +24,11 @@ type Bundle struct {
 	Errors  []string  `json:"errors,omitempty"`
 }
 
+// IsFinished returns true if the bundle has a status that would indicate that it is finished being created.
```
armandgrillet: `IsFinished returns if the bundle has a status that indicating that it is finished.`

br-lewis:
Squashed commits and fixed merge conflicts

bamarni (approved):

PR #345, opened 2019-08-28
Allow diagnostics download to default to the latest created bundle

br-lewis:
This also moves `latestBundle` to `pkg/cmd/diagnostics/diagnostics.go` because it's now used in multiple commands.

Addresses: https://jira.mesosphere.com/browse/DCOS_OSS-5489

### DIFF
--- a/pkg/cmd/diagnostics/diagnostics_download.go
+++ b/pkg/cmd/diagnostics/diagnostics_download.go
@@ -17,12 +17,21 @@ func newDiagnosticsDownloadCommand(ctx api.Context) *cobra.Command {
 		// the <bundle-id> is used to indicate what the expected positional arg is in the help output
 		Use:   "download <bundle-id>",
 		Short: "Download diagnostics bundle",
-		Args:  cobra.ExactArgs(1),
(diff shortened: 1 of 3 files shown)


janisz (approved):

pkg/cmd/diagnostics/diagnostics_download.go
```diff
@@ -17,12 +17,21 @@ func newDiagnosticsDownloadCommand(ctx api.Context) *cobra.Command {
...
 
-			id := args[0]
+			var id string
+			if len(args) == 0 {
+				bundle, err := latestBundle(client)
+				if err != nil {
+					return err
+				}
+				id = bundle.ID
+			} else {
```
drewvanstone, line 32: Don't need an else statement here

br-lewis: But then we'd always try to use `args[0]` which would crash if there are no arguments given.

drewvanstone (approved):

bamarni (approved):

PR #346, opened 2019-09-02
Refactor the rest of `dcos package repo...` into Go

br-lewis:
This changes `add`, `list`, and `remove` to be done in Go instead of calling out to the python CLI.
Addresses: https://jira.mesosphere.com/browse/DCOS_OSS-4776

### DIFF
--- a/pkg/cmd/pkg/package_repo_add.go
+++ b/pkg/cmd/pkg/package_repo_add.go
@@ -23,7 +23,11 @@ func newCmdPackageRepoAdd(ctx api.Context) *cobra.Command {
 				return err
 			}
 
-			_, err = c.PackageAddRepo(repoName, repoURL, index)
+			if cmd.Flags().Changed("index") {
+				_, err = c.PackageAddRepo(repoName, repoURL, &index)
+				return err
+			}
+			_, err = c.PackageAddRepo(repoName, repoURL, nil)
 			return err
 		},
 	}
--- a/pkg/cmd/pkg/package_repo_add.go
+++ b/pkg/cmd/pkg/package_repo_add.go
@@ -2,6 +2,8 @@ package pkg
 
 import (
 	"github.com/dcos/dcos-cli/api"
+	"github.com/dcos/dcos-core-cli/pkg/cosmos"
+	"github.com/dcos/dcos-core-cli/pkg/pluginutil"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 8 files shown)


makkes (approved):

pkg/cmd/pkg/package_repo_add.go
```diff
@@ -13,7 +15,16 @@ func newCmdPackageRepoAdd(ctx api.Context) *cobra.Command {
...
-			return invokePythonCLI(ctx)
+			repoName := args[0]
+			repoURL := args[1]
+
+			c, err := cosmos.NewClient(ctx, pluginutil.HTTPClient(""))
+			if err != nil {
+				return err
+			}
+
+			_, err = c.PackageAddRepo(repoName, repoURL, index)
```
bamarni: We need to make the distinction between passing `--index=0` and not passing `--index` at all. See https://docs.d2iq.com/mesosphere/dcos/1.13/deploying-services/package-api/#/default/POST_package_repository_add : 

> To add a package repository to the beginning of the list set the index to zero (0). To add a package repository to the end of the list do not specify an index.

But here we'd always be passing 0. The `PackageAddRepo()` method needs to take a pointer for the index. To check if the flag is passed with 0 or not passed at all we can use https://godoc.org/github.com/spf13/pflag#FlagSet.Changed

pkg/cmd/pkg/package_repo_list.go
```diff
@@ -18,7 +26,37 @@ func newCmdPackageRepoList(ctx api.Context) *cobra.Command {
...
+				return err
+			}
+
+			if jsonOutput {
+				enc := json.NewEncoder(ctx.Out())
+				enc.SetIndent("", "    ")
+				return enc.Encode(repos)
+			}
+
+			// TODO: this follows the python behavior but is this really an error?
```
bamarni: I'm pretty sure we had this case already while refactoring but don't remember about our decision, I'll try to search tomorrow on which `list` command it was. @rgo3 any suggestion?

bamarni: Looking at other commands like `dcos task list`, `dcos job list`, `dcos cluster list`, etc. We don't fail when there is no entries, but rather only print the table header. We should do the same for consistency.

bamarni (approved):

bamarni:

br-lewis:

janisz (approved):

br-lewis:

br-lewis:

br-lewis:

br-lewis:

br-lewis:

br-lewis:

br-lewis:

br-lewis:
timeout in the test, restarting without code changes this time.

br-lewis:

br-lewis:

PR #349, opened 2019-09-03
Fix makefile and add linters

janisz:


pkg/pluginutil/pluginutil.go
```diff
@@ -35,7 +35,9 @@ func HTTPClient(baseURL string, opts ...httpclient.Option) *httpclient.Client {
 
 	tlsInsecure, _ := os.LookupEnv("DCOS_TLS_INSECURE")
 	if tlsInsecure == "1" {
-		baseOpts = append(baseOpts, httpclient.TLS(&tls.Config{InsecureSkipVerify: true}))
+		baseOpts = append(baseOpts, httpclient.TLS(&tls.Config{
+			InsecureSkipVerify: true}, //nolint: gosec
```
bamarni: Looks a bit odd with the closing bracket like this, maybe more readable to have `//nolint: gosec` on top of `if tlsInsecure == "1" {`

pkg/mesos/taskio.go
```diff
@@ -510,7 +512,7 @@ func (t *TaskIO) ttyInfo(fd int) (*mesos.TTYInfo, error) {
 
 // forwardContainerOutput forwards output of the LAUNCH_NESTED_CONTAINER_SESSION
 // or ATTACH_CONTAINER_OUTPUT responses to STDOUT/STDERR.
-func (t *TaskIO) forwardContainerOutput(resp mesos.Response) error {
+func (t *TaskIO) forwardContainerOutput(resp encoding.Decoder) error {
```
bamarni, line 515: What is the rule exactly here? Is it because we only use methods from the embed interface?

pkg/cmd/quota/quota_list.go
```diff
@@ -41,20 +41,19 @@ func newCmdQuotaList(ctx api.Context) *cobra.Command {
...
-				return rolesResult.err
+			rResult := <-rolesRes
+			if rResult.err != nil {
+				return rResult.err
 			}
 
-			groupsResult := <-groupsRes
-			if groupsResult.err != nil {
-				return groupsResult.err
+			gResult := <-groupsRes
```
bamarni: What is the rule for not being able to use `groupsResult`?

pkg/cmd/job/job_show_runs.go
```diff
@@ -27,9 +27,9 @@ func newCmdJobShowRuns(ctx api.Context) *cobra.Command {
 
 			var runs []metronome.Run
 			if runID != "" {
-				run, err := client.Run(args[0], runID)
-				if err != nil {
-					return err
+				run, runErr := client.Run(args[0], runID)
```
bamarni: Why can't we use `err` here? Looks like a false-positive with variable shadowing. Especially for errors I think we should default to the idiomatic `err`, and only have different variable names when they need to be kept for later usage, which should be egde-cases.

pkg/cmd/task/task_list.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const notAvailable = "N/A"
```
bamarni: nit: it sounds arbitrary to have the constant in this `dcos task list` while it is used in other commands as well. Maybe have it in the root task command.

bamarni (commented):
Looks great in general, it catches more issues than what we are using currently. I left some suggestions/questions about a few changes.

pkg/mesos/taskio.go
janisz: Use interface instead of direct type

pkg/cmd/job/job_show_runs.go
janisz: It shadows `err` from above

pkg/cmd/quota/quota_list.go
janisz: I shadows declaration from above. I'm not sure why

br-lewis (approved):

pkg/cmd/diagnostics/diagnostics_download.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func newDiagnosticsDownloadCommand(ctx api.Context) *cobra.Command {
+func newDiagnosticsDownloadCommand(_ api.Context) *cobra.Command {
```
makkes: Why not remove this parameter completely when we're not using it?

cmd/dcos/main.go
```diff
@@ -18,7 +18,7 @@ func main() {
 	}
 }
 
-func run(ctx api.Context, args []string) error {
+func run(ctx api.Context, _ []string) error {
```
makkes: Why not remove this parameter completely when we're not using it?

pkg/cmd/node/node_diagnostics_download.go
```diff
@@ -23,9 +23,9 @@ func newCmdNodeDiagnosticsDownload(ctx api.Context) *cobra.Command {
 			}
 
 			if location == "" {
-				var err error
-				location, err = os.Getwd()
-				if err != nil {
+				var e error
```
makkes:
```suggestion
				var getwdErr error
```

pkg/cmd/node/node_list.go
```diff
@@ -53,24 +53,24 @@ func newCmdNodeList(ctx api.Context) *cobra.Command {
 			// results in a significant speedup of the command.
 			stateRes := make(chan stateResult)
 			go func() {
-				state, err := client.State()
-				stateRes <- stateResult{state, err}
+				state, e := client.State()
```
makkes:
```suggestion
				state, stateErr := client.State()
```

pkg/cmd/node/node_list.go
```diff
@@ -53,24 +53,24 @@ func newCmdNodeList(ctx api.Context) *cobra.Command {
...
+				stateRes <- stateResult{state, e}
 			}()
 
 			stateSummaryRes := make(chan stateSummaryResult)
 			go func() {
-				stateSummary, err := client.StateSummary()
-				stateSummaryRes <- stateSummaryResult{stateSummary, err}
+				stateSummary, e := client.StateSummary()
```
makkes:
```suggestion
				stateSummary, summaryErr := client.StateSummary()
```

pkg/cmd/node/node_list.go
```diff
@@ -53,24 +53,24 @@ func newCmdNodeList(ctx api.Context) *cobra.Command {
...
+				stateSummaryRes <- stateSummaryResult{stateSummary, e}
 			}()
 
 			ipsRes := make(chan ipsResult)
 			go func() {
 				c := networking.NewClient(pluginutil.HTTPClient(""))
-				nodes, err := c.Nodes()
+				nodes, e := c.Nodes()
```
makkes:
```suggestion
				nodes, nodesErr := c.Nodes()
```

pkg/cmd/node/node_list.go
```diff
@@ -81,8 +81,8 @@ func newCmdNodeList(ctx api.Context) *cobra.Command {
 
 			agentsRes := make(chan agentsResult)
 			go func() {
-				agents, err := client.Agents()
-				agentsRes <- agentsResult{agents, err}
+				agents, e := client.Agents()
```
makkes:
```suggestion
				agents, agentsErr := client.Agents()
```

pkg/cmd/quota/quota_create.go
```diff
@@ -52,21 +52,21 @@ func newCmdQuotaCreate(ctx api.Context) *cobra.Command {
 				rolesRes <- rolesResult{roles, err}
 			}()
 
-			rolesResult := <-rolesRes
-			if rolesResult.err != nil {
-				return rolesResult.err
+			rResult := <-rolesRes
```
makkes: I don't think this makes this variable name more expressive. I would rather rename the channel to `rolesResCh` or `rolesResChan`.

pkg/cmd/quota/quota_create.go
```diff
@@ -52,21 +52,21 @@ func newCmdQuotaCreate(ctx api.Context) *cobra.Command {
...
+			if rResult.err != nil {
+				return rResult.err
 			}
 
-			groupsResult := <-groupsRes
-			if groupsResult.err != nil {
-				return groupsResult.err
+			gResult := <-groupsRes
```
makkes: same here

pkg/cmd/quota/quota_list.go
makkes: see my comments above. It shadows the type decl in pkg/cmd/quota/quota.go

pkg/cmd/service/service_log.go
```diff
@@ -51,9 +51,9 @@ func newCmdServiceLog(ctx api.Context) *cobra.Command {
 			}
 
 			if args[0] == "marathon" {
-				url, err := marathonMasterURL(marathonClient, ctx)
-				if err != nil {
-					return err
+				url, e := marathonMasterURL(marathonClient, ctx)
```
makkes:
```suggestion
				url, urlErr := marathonMasterURL(marathonClient, ctx)
```

makkes (changes requested):

pkg/mesos/taskio.go
bamarni: I don't get it, `mesos.Response` is an interface already

janisz:
```
pkg/mesos/taskio.go:513:41: `resp` can be `github.com/mesos/mesos-go/api/v1/lib/encoding.Decoder` (interfacer)
func (t *TaskIO) forwardContainerOutput(resp mesos.Response) error {
                                        ^
```

We are only using `Decode` method so there is no need to use `mesos.Response` and we can narrow it to `encoding.Decoder`

pkg/cmd/job/job_show_runs.go
bamarni: My point is that it's a false-positive. As errors are usually checked only once and right after they're declared, it is fine to have `err` here and I think it's idiomatic Go.

Even when I vet the Go source code itself for shadowing there are hundreds of errors like `src/archive/tar/reader.go:81:6: declaration of "err" shadows declaration at line 77`.

pkg/mesos/taskio.go
bamarni: I see, in this specific case it should be fine, though I have mixed feelings about this rule in general:

- if there is an interface which is exposed in a package as the return type, maybe the embed interface is an implementation detail that might change, or eg. it was an old interface which is getting deprecated for the one explicitly stated.

- does the rule also applies to public methods? For example what if `ForwardContainerOutput(mesos.Response) error` was part of an interface I wanted to implement for `TaskIO`.

janisz: >  maybe the embed interface is an implementation detail that might change

It's not possible, embeded interface is part of interface itself.

> does the rule also applies to public methods? For example what if ForwardContainerOutput(mesos.Response) error was part of an interface I wanted to implement for TaskIO

If there is an interface you want to implement then error will not be triggered

pkg/cmd/service/service_log.go
janisz: Fixed

pkg/cmd/node/node_list.go
janisz: Fixed

pkg/cmd/node/node_list.go
janisz: Fixed

pkg/cmd/node/node_list.go
janisz: Fixed

pkg/cmd/node/node_list.go
janisz: Fixed

pkg/cmd/node/node_diagnostics_download.go
janisz: Fixed

cmd/dcos/main.go
janisz: Fixed

pkg/cmd/diagnostics/diagnostics_download.go
janisz: Fixed

pkg/cmd/task/task_list.go
janisz: Fixed

pkg/cmd/job/job_show_runs.go
janisz: > My point is that it's a false-positive.

How it could be a false positive if it shadows previous declaration

> Even when I vet the Go source code itself for shadowing there are hundreds

Go code if full of some practices we do not follow. Take a look at @makkes comments on `e` error name. Go code is full of single variable names, or weird 3 letters words that are hard to decrypt. Do we also want to follow that rule? If you do not like this check we can disable it but IMO it's better to have errors with clear names indicating what they are used for.

makkes (approved):

pkg/cmd/job/job_show_runs.go
bamarni: > How it could be a false positive if it shadows previous declaration

It is a false positive because the `err` variable of the outer block is not meant to be changed here.

Furthermore, the outer `err` variable is referenced below in the else block only, so I'd even argue that it is shadowing at all here, there is no code path going through the inner and outer `err` variable afterwards.

I'd rather remove this rule, see https://github.com/golang/go/issues/19490#issuecomment-285581930 where they say it is known to give many false-positives and still experimental.

janisz: I disabled shadow check

bamarni (approved):

Makefile
```diff
@@ -34,17 +34,12 @@ python:
...
 
-.PHONY: vet
-vet: lint
-	$(call inDocker,go vet ./...)
-
 .PHONY: lint
 lint: docker-image
-	# Can be simplified once https://github.com/golang/lint/issues/320 is fixed.
-	$(call inDocker,golint -set_exit_status ./cmd/... ./pkg/...)
+	$(call inDocker, golangci-lint -v run)
```
bamarni:
```suggestion
	$(call inDocker, golangci-lint run)
```

I noticed when running locally that it's very verbose even when there is no error, other commands like `go vet` or `golint` don't print anything on success.

PR #357, opened 2019-09-17
[2.0] Add missing fields in job definition

bamarni:
The Go types were not in sync with the actual definition.

https://github.com/dcos/dcos-cli/issues/1494

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## 2.0-patch.x
+
+* Fixes
+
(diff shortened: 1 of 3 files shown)


janisz:
How about generating this structs from [Marathon](https://github.com/mesosphere/marathon/blob/master/docs/docs/rest-api/public/api/api.raml) and [Metronome RAML API definiton](https://github.com/dcos/metronome/blob/master/api/src/main/resources/public/api/api.raml) with https://github.com/Jumpscale/go-raml

makkes (approved):

PR #358, opened 2019-09-17
Refactor dcos package install in Go

janisz:
Fixes: https://jira.mesosphere.com/browse/DCOS_OSS-4774


janisz:

janisz:

pkg/cosmos/client.go
```diff
@@ -156,6 +159,30 @@ func (c *Client) PackageRender(appID string, name string, version string, option
 	return render.MarathonJson, nil
 }
 
+// PackageInstall installs package
+func (c *Client) PackageInstalls(appID string, name string, version string, optionsPath string) error {
```
drewvanstone: I'm not sure I understand why this is called `PackageInstalls` vs `PackageInstall`

pkg/cmd/pkg/package_install.go
```diff
@@ -1,26 +1,160 @@
...
 
 	cmd := &cobra.Command{
 		Use:   "install <package-name>",
 		Short: "Install a package",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return invokePythonCLI(ctx)
+			packageName := args[0]
+
+			if cliOnly == appOnly {
```
makkes: So if both are `true` or `false` we set both to `true`? I think at least a comment would be helpful here for later reference.

bamarni: Btw, passing both `--cli` and `--app` is currently not allowed, shall we do the same here?

pkg/cmd/pkg/package_install.go
```diff
@@ -1,26 +1,160 @@
...
+					"There may be bugs, incomplete features, incorrect documentation, or other discrepancies.\n")
+				link = "https://mesosphere.com/catalog-terms-conditions/#community-services"
+			}
+			fmt.Fprintf(ctx.Out(), "By Deploying, you agree to the Terms and Conditions %s\n", link)
+			if appOnly && pkg.PreInstallNotes != "" {
+				fmt.Fprintf(ctx.Out(), "%s\n", pkg.PreInstallNotes)
+			}
+
+			if !yes {
+				prompter := prompt.New(ctx.Input(), ctx.Out())
```
bamarni:
```suggestion
				prompter := ctx.Prompt()
```

pkg/cmd/pkg/package_install.go
```diff
@@ -1,26 +1,160 @@
...
+				return err
+			}
+
+			description, err := c.PackageDescribe(packageName, version)
+			if err != nil {
+				return err
+			}
+			pkg := description.Package
+
+			link := "https://mesosphere.com/catalog-terms-conditions/#certified-services"
```
makkes: I get a 404 for this URL.

pkg/cmd/pkg/package_install.go
```diff
@@ -1,26 +1,160 @@
...
+			if !pkg.Selected {
+				fmt.Fprint(ctx.Out(), "This is a Community service. "+
+					"Community services are not tested for production environments. "+
+					"There may be bugs, incomplete features, incorrect documentation, or other discrepancies.\n")
+				link = "https://mesosphere.com/catalog-terms-conditions/#community-services"
```
makkes: Also 404.

pkg/cmd/pkg/package_install.go
```diff
@@ -1,26 +1,160 @@
...
+
+			description, err := c.PackageDescribe(packageName, version)
+			if err != nil {
+				return err
+			}
+			pkg := description.Package
+
+			link := "https://mesosphere.com/catalog-terms-conditions/#certified-services"
+			if !pkg.Selected {
+				fmt.Fprint(ctx.Out(), "This is a Community service. "+
```
bamarni: These informational messages should go to stderr according to https://github.com/dcos/dcos-cli/blob/master/design/style.md#stdoutstderr, I think this refactoring is a good time to switch from stdout to stderr.

pkg/cmd/pkg/package_install.go
```diff
@@ -1,26 +1,160 @@
...
+				fmt.Fprintf(ctx.Out(), "Installing CLI subcommand for package [%s] version [%s]\n", packageName, pkg.Version)
+				pluginInfo, err := cosmos.CLIPluginInfo(description.Package, pluginutil.HTTPClient("").BaseURL())
+				if err != nil {
+					return fmt.Errorf("cannot get plugin info: %s", err)
+				}
+
+				var checksum plugin.Checksum
+				for _, contentHash := range pluginInfo.ContentHash {
+					switch contentHash.Algo {
+					case dcos.SHA256:
```
makkes: What's with the other cases? I think a comment explaining what's going on here would help as well.

pkg/cosmos/client.go
```diff
@@ -223,3 +250,34 @@ func cosmosErrUnwrap(err error) error {
 		return err
 	}
 }
+
+// CLIPluginInfo extracts plugin resource data from the Cosmos package and for the current platform.
+func CLIPluginInfo(pkg dcos.CosmosPackage, baseURL *url.URL) (cliArtifact dcos.CosmosPackageResourceCliArtifact, err error) {
+	switch runtime.GOOS {
```
makkes, line 267: I'm missing a default case here.

pkg/cmd/pkg/package_install.go
makkes: I'm also missing a default case here.

pkg/cmd/pkg/package_install.go
```diff
@@ -1,26 +1,160 @@
...
+						return json.NewEncoder(pkgInfoFile).Encode(description.Package)
+					},
+				})
+				if err != nil {
+					return fmt.Errorf("cannot install plugin: %s", err)
+				}
+				// See: https://github.com/dcos/dcos-cli/blob/72953d7eb2821f966e823294c4a451b071ee0f29/pkg/plugin/manager.go#L417
+				pluginName := packageName
+				if pluginInfo.Kind == "zip" {
+					pluginName = "dcos-" + pluginName
```
bamarni: Why adding this condition? The plugin name should always be packageName, as we pass it in line 111.

makkes (changes requested):

pkg/cmd/pkg/package_install.go
```diff
@@ -1,26 +1,160 @@
...
+				plural := ""
+				if len(plugin.Commands) > 1 {
+					plural = "s"
+				}
+				cmds := make([]string, 0, len(plugin.Commands))
+				for _, c := range plugin.Commands {
+					cmds = append(cmds, c.Name)
+				}
+
+				fmt.Fprintf(ctx.Out(), "New command%s available: dcos %s\n", plural, strings.Join(cmds, ", "))
```
bamarni:
```suggestion
				sort.Strings(cmds)
				fmt.Fprintf(ctx.ErrOut(), "New command%s available: dcos %s\n", plural, strings.Join(cmds, ", "))
```
Even though usually there is a single command, it'd be nice to enforce sorting here.

pkg/cmd/pkg/package_install.go
```diff
@@ -1,26 +1,160 @@
...
+			if appOnly && pkg.PostInstallNotes != "" {
+				fmt.Fprintf(ctx.Out(), "%s\n", pkg.PostInstallNotes)
+			}
+			return nil
 		},
 	}
 	cmd.Flags().BoolVar(&appOnly, "app", false, "Application only")
 	cmd.Flags().StringVar(&appID, "app-id", "", "The application ID")
 	cmd.Flags().BoolVar(&cliOnly, "cli", false, "Command line only")
-	cmd.Flags().BoolVar(&global, "global", false, "Install a subcommand for all configured clusters")
```
bamarni, line 23: :+1:

pkg/cmd/pkg/package_install.go
janisz: So you are the first person who tried to visit that page :)
I'm not sure what to do about it :(
@drewvanstone what messages should we use here instead?

pkg/cmd/pkg/package_install.go
```diff
@@ -1,26 +1,160 @@
...
-			return invokePythonCLI(ctx)
+			packageName := args[0]
+
+			if cliOnly == appOnly {
+				cliOnly = true
+				appOnly = true
+			}
+
+			if optionsPath != "" {
+				_, err := os.Stat(optionsPath)
```
bamarni: nit: maybe simpler not to have stat and let the read fail later on.

pkg/cmd/pkg/package_install.go
bamarni: Maybe was dropped during website changes for d2iq

pkg/cmd/pkg/package_install.go
bamarni: We only support this algorithm at the moment (https://github.com/dcos/cosmos/blob/3109d65d2e4abb66aad9b30ca9c5cb6e90469e63/cosmos-common/src/main/resources/com/mesosphere/universe/v3/schema/repo#L60).

But that's a good point that we might want to do defensive programming here and prepare for new algorithms.

I think failing would be too aggressive, what about adding a warning if the algo is unknown and skipping checksum validation?

pkg/cmd/pkg/package_install.go
janisz: https://mesosphere.slack.com/archives/C02526BKZ/p1568810254000500

makkes: This is not blocking this PR, rather a follow-up item.

pkg/cmd/pkg/package_install.go
janisz: I think we should fail when new algorithm appears

pkg/cmd/pkg/package_install.go
janisz: I think it's fine to pass both `--cli --app` this is more explicit.

pkg/cmd/pkg/package_install.go
janisz: It's not true. Plugin name depends on plugin kind see above comment

pkg/cmd/pkg/package_install.go
janisz: Then I'll need to change test that checks if we fail fast on wrong config

pkg/cosmos/client.go
janisz: Fixed

pkg/cmd/pkg/package_install.go
janisz: Fixed

pkg/cmd/pkg/package_install.go
janisz: Fixed

pkg/cmd/pkg/package_install.go
janisz: Fixed

pkg/cmd/pkg/package_install.go
janisz: Fixed

pkg/cmd/pkg/package_install.go
```diff
@@ -1,26 +1,162 @@
...
 
 	cmd := &cobra.Command{
 		Use:   "install <package-name>",
 		Short: "Install a package",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return invokePythonCLI(ctx)
+			packageName := args[0]
+
+			// Install both if neither flag is specified
```
makkes: I still don't get this, please bear with me. This check is true when either `cliOnly` and `appOnly` are true or both are false. This comment suggests it's only when both are false. I'd like to understand the reasoning behind this if.

janisz:
Should I add a unit test?

makkes:
> Should I add a unit test?

I
Would
Love
That!

pkg/cmd/pkg/package_install.go
janisz: We should wait for https://github.com/dcos/dcos-cli/pull/1497

pkg/cmd/pkg/package_install.go
bamarni: Passing `--app` means: `Only install the package on the cluster side but I don't want the CLI subcommand locally`.

Passing `--cli` means: `Skip the package installation and only install the CLI subcommand locally`

Passing both `--cli` and `--app` is not supported currently and returns with an error, these options are mutually exclusive, so having both `cliOnly` and `appOnly` should not happen, even though it is not enforced in this PR.

makkes: >  both --cli and --app is not supported currently and returns with an error

I can't find the code in this PR that enforces this behaviour.

pkg/cmd/pkg/package_install.go
king724: The URL has been fixed.

pkg/cmd/pkg/package_install.go
bamarni: By "currently", I meant with the Python command, this is not enforced anymore with this PR.

janisz: Fixed

pkg/cmd/pkg/package_install.go
janisz: Done

janisz:

janisz:

pkg/cmd/pkg/package_install.go
```diff
@@ -1,28 +1,163 @@
...
+
+		plural := ""
+		if len(plugin.Commands) > 1 {
+			plural = "s"
+		}
+		cmds := make([]string, 0, len(plugin.Commands))
+		for _, c := range plugin.Commands {
+			cmds = append(cmds, c.Name)
+		}
+		sort.Strings(cmds)
```
bamarni: You could as well use the `plugin.CommandNames()` helper to get a sorted list of commands.

janisz: Fixed

janisz:

janisz:

janisz:
I accidentally closed this PR. Since I can't re-open it I opened a new one https://github.com/dcos/dcos-core-cli/pull/361

PR #360, opened 2019-09-20
Add tests for describe

janisz:
Example how we can move some of our test from E2E python tests into faster Go unit tests.

### DIFF
--- a/pkg/cmd/pkg/package_describe.go
+++ b/pkg/cmd/pkg/package_describe.go
@@ -12,82 +12,89 @@ import (
 	"github.com/dcos/dcos-core-cli/pkg/pluginutil"
 )
 
-func newCmdPackageDescribe(ctx api.Context) *cobra.Command {
-	var allVersions, appOnly, cliOnly, config, render bool
-	var appID, optionsPath, version string
+type describeOptions struct {
+	appID       string
+	optionsPath string
+	version     string
 
+	allVersions bool
+	appOnly     bool
+	cliOnly     bool
+	config      bool
+	render      bool
+}
+
+func newCmdPackageDescribe(ctx api.Context) *cobra.Command {
+	var opts describeOptions
 	cmd := &cobra.Command{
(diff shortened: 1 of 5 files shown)


pkg/cmd/pkg/package_describe.go
```diff
@@ -12,82 +12,89 @@ import (
...
+	cmd.Flags().StringVar(&opts.version, "package-opts.version", "", "The package opts.version")
+	cmd.Flags().BoolVar(&opts.allVersions, "package-versions", false, " Displays all versions for this package")
+	cmd.Flags().BoolVar(&opts.render, "opts.render", false,
+		"Collate the `marathon.json` package template with values from the `opts.config.json` and `--describeOptions`")
+	return cmd
+}
 
-			}
+func pkgDescribe(ctx api.Context, packageName string, opts describeOptions) error {
+	c, err := cosmos.NewClient(ctx, pluginutil.HTTPClient(""))
```
makkes, line 50: Injecting this dependency would make it easier (or even possible in the first place) to unit test this func. We should then test the `cosmos.Client` independently from this command.

pkg/cmd/pkg/package_describe_test.go
```diff
@@ -0,0 +1,72 @@
...
+func Test_pkgDescribe(t *testing.T) {
+	var out bytes.Buffer
+	env := mock.NewEnvironment()
+	env.Fs = afero.NewCopyOnWriteFs(
+		afero.NewReadOnlyFs(afero.NewOsFs()),
+		afero.NewMemMapFs(),
+	)
+	env.Out = &out
+
+	testServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
makkes, line 27: If we injected `comsos.Client` into the command func we wouldn't have the burden of spinning up this server and the test would be much more readable.

makkes (commented):
Very nice! Just some remarks.

PR #361, opened 2019-09-23
[DCOS_OSS-4774] Refactor 'dcos package install' in Go

janisz:
**Fixes:** [DCOS_OSS-4774](https://jira.mesosphere.com/browse/DCOS_OSS-4774)
Refs: https://github.com/dcos/dcos-core-cli/pull/358


br-lewis (approved):

janisz:

pkg/cmd/pkg/package_install.go
```diff
@@ -1,28 +1,159 @@
...
+func pkgInstall(ctx api.Context, packageName string, opts pkgInstallOptions) error {
+	if packageName == "" {
+		return fmt.Errorf("package name must not be empty")
+	}
+
+	if opts.cliOnly && opts.appOnly {
+		return fmt.Errorf("--app and --cli are mutually exclusive")
+	}
+	// Install both if neither flag is specified
+	if opts.cliOnly == opts.appOnly {
```
makkes, line 59: So a user cannot provide both flags but when he provides none it's implied he provided both?

makkes (commented):
Vendoring really makes code reviews much harder...

pkg/cmd/pkg/package_install.go
janisz: Correct, @bamarni can you explain more?

bamarni: > when he provides none it's implied he provided both

I've left a comment about this in https://github.com/dcos/dcos-core-cli/pull/358#discussion_r326216600. Another way to looks at this (and I think it would have been clearer), is that `--cli` was called `--no-app` and `--app` was called `--no-cli`. Passing none of them means they want to install both (the app and the CLI). Passing both of them means "install nothing", so it's not really valid.

bamarni: Maybe the code would be less confusing without this if condition. It can be done eg. by having `installCLI` and `installPackage` as installation options instead of `cliOnly` and `appOnly`, where `installCLI` and `installPackage` are set individually to `!appOnly` and `!cliOnly`.

janisz: Can you provide an example. It's hard fro me to think about all this flags in negative way

bamarni:
``` diff
diff --git a/pkg/cmd/pkg/package_install.go b/pkg/cmd/pkg/package_install.go
index a30fc442..a7809953 100644
--- a/pkg/cmd/pkg/package_install.go
+++ b/pkg/cmd/pkg/package_install.go
@@ -22,8 +22,8 @@ type pkgInstallOptions struct {
 	appID       string
 	optionsPath string
 	version     string
-	appOnly     bool
-	cliOnly     bool
+	app         bool
+	cli         bool
 	yes         bool
 }
 
@@ -38,9 +38,9 @@ func newCmdPackageInstall(ctx api.Context) *cobra.Command {
 			return pkgInstall(ctx, args[0], options)
 		},
 	}
-	cmd.Flags().BoolVar(&options.appOnly, "app", false, "Application only")
+	cmd.Flags().BoolVar(&options.app, "app", false, "Application only")
 	cmd.Flags().StringVar(&options.appID, "app-id", "", "The application ID")
-	cmd.Flags().BoolVar(&options.cliOnly, "cli", false, "Command line only")
+	cmd.Flags().BoolVar(&options.cli, "cli", false, "Command line only")
 	cmd.Flags().StringVar(&options.optionsPath, "options", "", "Path to a JSON file that contains customized package installation options")
 	cmd.Flags().StringVar(&options.version, "package-version", "", "The package version")
 	cmd.Flags().BoolVar(&options.yes, "yes", false, "Disable interactive mode and assume “yes” is the answer to all prompts")
@@ -52,14 +52,11 @@ func pkgInstall(ctx api.Context, packageName string, opts pkgInstallOptions) err
 		return fmt.Errorf("package name must not be empty")
 	}
 
-	if opts.cliOnly && opts.appOnly {
+	if opts.app && opts.cli {
 		return fmt.Errorf("--app and --cli are mutually exclusive")
 	}
-	// Install both if neither flag is specified
-	if opts.cliOnly == opts.appOnly {
-		opts.cliOnly = true
-		opts.appOnly = true
-	}
+	opts.app, opts.cli = !opts.cli, !opts.app
+
 	if opts.optionsPath != "" {
 		_, err := os.Stat(opts.optionsPath)
 		if err != nil {
@@ -83,7 +80,7 @@ func pkgInstall(ctx api.Context, packageName string, opts pkgInstallOptions) err
 		link = "https://mesosphere.com/catalog-terms-conditions/#community-services"
 	}
 	fmt.Fprintf(ctx.ErrOut(), "By Deploying, you agree to the Terms and Conditions %s\n", link)
-	if opts.appOnly && pkg.PreInstallNotes != "" {
+	if opts.app && pkg.PreInstallNotes != "" {
 		fmt.Fprintf(ctx.ErrOut(), "%s\n", pkg.PreInstallNotes)
 	}
 	if !opts.yes {
@@ -92,18 +89,18 @@ func pkgInstall(ctx api.Context, packageName string, opts pkgInstallOptions) err
 			return err
 		}
 	}
-	if opts.appOnly && description.Package.Marathon.V2AppMustacheTemplate != "" {
+	if opts.app && description.Package.Marathon.V2AppMustacheTemplate != "" {
 		fmt.Fprintf(ctx.ErrOut(), "Installing Marathon app for package [%s] version [%s]\n", packageName, pkg.Version)
 		err := c.PackageInstall(opts.appID, packageName, pkg.Version, opts.optionsPath)
 		if err != nil {
 			return err
 		}
 	}
-	if opts.cliOnly && isEmptyCli(pkg.Resource.Cli) && !isEmptyCommand(pkg.Command) {
+	if opts.cli && isEmptyCli(pkg.Resource.Cli) && !isEmptyCommand(pkg.Command) {
 		return fmt.Errorf("unable to install CLI subcommand. PIP subcommands are no longer supported" +
 			" see: https://godoc.org/github.com/dcos/client-go/dcos#CosmosPackageCommand")
 	}
-	if opts.cliOnly && !isEmptyCli(pkg.Resource.Cli) {
+	if opts.cli && !isEmptyCli(pkg.Resource.Cli) {
 		fmt.Fprintf(ctx.ErrOut(), "Installing CLI subcommand for package [%s] version [%s]\n", packageName, pkg.Version)
 		pluginInfo, err := cosmos.CLIPluginInfo(description.Package, pluginutil.HTTPClient("").BaseURL())
 		if err != nil {
@@ -152,7 +149,7 @@ func pkgInstall(ctx api.Context, packageName string, opts pkgInstallOptions) err
 
 		fmt.Fprintf(ctx.ErrOut(), "New command%s available: dcos %s\n", plural, strings.Join(cmds, ", "))
 	}
-	if opts.appOnly && pkg.PostInstallNotes != "" {
+	if opts.app && pkg.PostInstallNotes != "" {
 		fmt.Fprintf(ctx.ErrOut(), "%s\n", pkg.PostInstallNotes)
 	}
 	return nil
```

bamarni: However In both cases it seems like what would look strange to someone looking at the code for the first time is that we change the semantics of the bool variables. Eg. at the beginning `appOnly` means `no CLI` and it's mutually exclusive with `cliOnly`, then we flip things and it means `install the app`, and it's not mutually exclusive with `cliOnly` anymore.

Maybe using 2 separate pair of variables would be clearer, but it's now getting into a nitpick :smiley_cat:

makkes (approved):
Ok, I think we can merge this since the logic that the code implies is actually the logic we want to expose. Whoever has a great idea of how to handle the `appOnly`/`cliOnly` madness is free to open a subsequent PR.

pkg/cmd/pkg/package_install.go
makkes: I still find it very strange that this is the logic we impose on the user:
```
appOnly | cliOnly | meaning
================================
false   | false   | install both
true    | false   | install app
false   | true    | install cli
true    | true    | error
```
Why not just make case 4 (true/true) the same as case 1 (false/false) and rename the flags as @bamarni suggested? Because now we're saying "no no no, you cannot provide both flags if you want to install the app AND the CLI. If this is what you want, provide none of those flags."

janisz: If we change meaning to install app and install cli and set it by default to true then there will be no problem

PR #322, opened 2019-08-15
Enforce role by default when creating a Marathon group

bamarni:

PR #366, opened 2019-10-03
Add test with multi level group

janisz:
This test should be failing due to Marahton API validation allowing only top-level group to be marked as `enforcedRole`. Reverting 1fcc664f908018969c2ea283be0a079a70a8a2ce fixed the issue

See:
* https://github.com/dcos/dcos-core-cli/pull/322
* https://github.com/mesosphere/marathon/pull/6989

Fixes: https://jira.mesosphere.com/browse/DCOS-59486


janisz:

janisz:

drewvanstone (approved):

python/lib/dcoscli/tests/integrations/test_marathon_groups.py
```diff
@@ -35,6 +35,11 @@ def test_add_existing_group():
                            stdin=fd)
 
 
+def test_add_non_root_group_from_stdin():
+    _add_group_by_stdin('tests/data/marathon/groups/simple.json')
+    remove_group('/x')
```
makkes, line 43: * What does this test? I see that a group is added and then removed but what are the constraints here?
* How does this test relate to the code changed above?

janisz: I added a comment

makkes: I don't think this setup is ideal. It'll break as soon as someone changes the `simple.json` but not this line. Since `simple.json` is only a single line I would just pass the literal data here so it's all in one place.

makkes (changes requested):

python/lib/dcoscli/dcoscli/marathon/main.py
```diff
@@ -488,10 +488,6 @@ def group_add(self, group_resource, id_):
         else:
             group_resource = self._resource_reader.get_resource(group_resource)
 
-        if 'enforceRole' not in group_resource:
-            # enforceRole by default, this is in-line with the UI.
-            group_resource['enforceRole'] = True
```
bamarni, line 493: This was added for consistency with the UI, they are setting it to true by default when creating a group. However I had missed that it's only applicable to top-level groups. It's still something we should do for top-level groups.

janisz:

python/lib/dcoscli/dcoscli/marathon/main.py
bamarni: However the behavior changed recently in DC/OS (https://jira.mesosphere.com/browse/DCOS_OSS-5532). So passing it explicitly might not be needed anymore for top-level groups, I'll check it out tomorrow on a dev DC/OS version.

janisz: Can you link to UI code and JIRA issue?
I think we should not apply our custom defaults. If this is a "good" default value then Marathon  should apply it. Imagine user who try to deploy application and try to debug why JSON was altered. There is no information in changelog about this change, nor test. We should remove this code.

cc. @meichstedt

bamarni: > Can you link to UI code and JIRA issue?
I think we should not apply our custom defaults. If this is a "good" default value then Marathon should apply it. 

See the issue I've pasted above, it is exactly about that. I've just tested against a 2.1-dev DC/OS cluster:

> dcos http -X POST -d '{"id": "/my-group"}' /service/marathon/v2/groups

And the group has `enforceRole` set to true by default.

```
  {
    "apps": [],
    "dependencies": [],
    "enforceRole": true,
    "groups": [],
    "id": "/my-group",
    "pods": [],
    "version": "2019-10-04T09:59:16.595Z"
  }
```

So we don't have to pass it explicitly anymore.

bamarni (approved):

python/lib/dcoscli/tests/integrations/test_marathon_groups.py
janisz: I agree but all our tests are using files and all helper functions are relying on file. If we would like to keep everything in this test I'd need to create temp file or create new helper that can handle string command stdin.

See:
* https://github.com/dcos/dcos-core-cli/blob/cdbbfce1579bf0c4d706bfb3675ec775d0f0713a/python/lib/dcoscli/tests/integrations/test_marathon_groups.py#L120
* https://github.com/dcos/dcos-core-cli/blob/55b3085ed15a1b26df8f0fbce58c15958c2541cc/python/lib/dcoscli/dcoscli/test/common.py#L24

janisz:
@makkes Can you re-review

makkes (approved):

PR #370, opened 2019-10-08
Allow bundle creation for master/agents only

janisz:
Previous API allow fine grained control over which nodes data should be included in the bundle. This change adds the most used function to create master only bundle with `dcos diagnostics create --agents=false`.


janisz:

PR #371, opened 2019-10-08
Allow bundle creation only for nodes or masters

janisz:
Previous API allow fine grained control over which nodes data should be
included in the bundle. This change adds the most used function to
create master only bundle with `dcos diagnostics create --masters`.

### DIFF
--- a/pkg/cmd/diagnostics/diagnostics_create.go
+++ b/pkg/cmd/diagnostics/diagnostics_create.go
@@ -10,6 +10,7 @@ import (
 )
 
 func newDiagnosticsCreateCommand(ctx api.Context) *cobra.Command {
+	opts := diagnostics.Options{}
 	cmd := &cobra.Command{
 		Use:   "create",
 		Short: "Create a diagnostics bundle",
@@ -17,7 +18,7 @@ func newDiagnosticsCreateCommand(ctx api.Context) *cobra.Command {
 			c := pluginutil.HTTPClient("")
 			client := diagnostics.NewClient(c)
 
-			id, err := client.Create()
+			id, err := client.Create(opts)
 			if err != nil {
 				return err
 			}
@@ -26,6 +27,8 @@ func newDiagnosticsCreateCommand(ctx api.Context) *cobra.Command {
 			return err
 		},
 	}
+	cmd.Flags().BoolVar(&opts.Masters, "masters", true, "Enable bundle collection on masters")
+	cmd.Flags().BoolVar(&opts.Agents, "agents", true, "Enable bundle collection on agents")
 
 	return cmd
 }
--- a/pkg/cmd/diagnostics/diagnostics_create.go
+++ b/pkg/cmd/diagnostics/diagnostics_create.go
@@ -10,14 +10,23 @@ import (
 )
 
 func newDiagnosticsCreateCommand(ctx api.Context) *cobra.Command {
-	opts := diagnostics.Options{}
+	var masters bool
 	cmd := &cobra.Command{
 		Use:   "create",
 		Short: "Create a diagnostics bundle",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			c := pluginutil.HTTPClient("")
 			client := diagnostics.NewClient(c)
-
+			opts := diagnostics.Options{
+				Masters: true,
+				Agents:  true,
+			}
+			if masters {
+				opts = diagnostics.Options{
+					Masters: true,
+					Agents:  false,
+				}
+			}
 			id, err := client.Create(opts)
 			if err != nil {
 				return err
@@ -27,8 +36,7 @@ func newDiagnosticsCreateCommand(ctx api.Context) *cobra.Command {
 			return err
 		},
 	}
-	cmd.Flags().BoolVar(&opts.Masters, "masters", true, "Enable bundle collection on masters")
-	cmd.Flags().BoolVar(&opts.Agents, "agents", true, "Enable bundle collection on agents")
+	cmd.Flags().BoolVar(&masters, "masters", true, "Collect data from masters only")
 
 	return cmd
 }
--- a/pkg/cmd/diagnostics/diagnostics_create.go
+++ b/pkg/cmd/diagnostics/diagnostics_create.go
@@ -10,7 +10,7 @@ import (
 )
 
 func newDiagnosticsCreateCommand(ctx api.Context) *cobra.Command {
-	var masters bool
+	var mastersOnly bool
 	cmd := &cobra.Command{
 		Use:   "create",
 		Short: "Create a diagnostics bundle",
@@ -19,13 +19,7 @@ func newDiagnosticsCreateCommand(ctx api.Context) *cobra.Command {
 			client := diagnostics.NewClient(c)
 			opts := diagnostics.Options{
 				Masters: true,
-				Agents:  true,
-			}
-			if masters {
-				opts = diagnostics.Options{
-					Masters: true,
-					Agents:  false,
-				}
+				Agents:  !mastersOnly,
 			}
 			id, err := client.Create(opts)
 			if err != nil {
@@ -36,7 +30,7 @@ func newDiagnosticsCreateCommand(ctx api.Context) *cobra.Command {
 			return err
 		},
 	}
-	cmd.Flags().BoolVar(&masters, "masters", true, "Collect data from masters only")
+	cmd.Flags().BoolVar(&mastersOnly, "masters", false, "Collect data from masters only")
 
 	return cmd
 }
--- a/pkg/cmd/diagnostics/diagnostics_create.go
+++ b/pkg/cmd/diagnostics/diagnostics_create.go
@@ -19,7 +19,7 @@ func newDiagnosticsCreateCommand(ctx api.Context) *cobra.Command {
 			client := diagnostics.NewClient(c)
 			opts := diagnostics.Options{
 				Masters: masters || !masters && !agents,
-				Agents: agents || !masters && !agents,
+				Agents:  agents || !masters && !agents,
 			}
 			id, err := client.Create(opts)
 			if err != nil {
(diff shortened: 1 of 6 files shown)


janisz:

janisz:

br-lewis (approved):

bamarni:
> This change adds the most used function to create master only bundle with dcos diagnostics create --agents=false

On the JIRA ticket it says `dcos --masters` to get masters only, were there any reason for reversing the behavior? From a UX perspective I think `--masters` is more intuitive than `--agents=false`.

We should also have a changelog entry for this new feature.

janisz:
I choos this implementation because I fell flags with only lead us to the same problems like we had with `appOnly` and `cliOnly` where by default both are true but when you specify one other is disabled. I'm not sure how to code it.

janisz:
@bamarni Is it better now?

janisz:

bamarni:
What seems hard to grasp here is that in order to select nodes with the diagnostics API, one has to negate other stuff instead of just picking what they want. For example, with the latest changes in the diagnostics API, all the following are equivalent and lead to a bundle with all the nodes:

- omitting the request body
- a request body of `{}`
- a request body of `{"masters": true}` (here one might expect to get only the masters)

Having to think about what we want to exclude rather than what we want seems to be backwards from how a user would think about this. To illustrate this even more, we had also talked about potentially adding flags to select specific nodes in the future. If eg. I want to get the masters and a given agent, I think `create --masters --node=my_node_1` is easier to get than `create --agents=false --node=my_node_1`, which mixes inclusion / exclusion.

If that's an option, we can probably make it easier for consumers at the diagnostics API level, by defaulting to true everywhere only when no explicit filter is passed.

janisz:
I removed `--agents` flag. Why you haven't raised you objections when we designed API and implemented it?

pkg/cmd/diagnostics/diagnostics_create.go
```diff
@@ -10,14 +10,24 @@ import (
...
-			id, err := client.Create()
+			opts := diagnostics.Options{
+				Masters: true,
+				Agents:  true,
+			}
+			if masters {
+				opts = diagnostics.Options{
+					Masters: true,
+					Agents:  false,
+				}
```
makkes:
```
opts.Agents = false
```

pkg/diagnostics/v2/client.go
```diff
@@ -86,8 +93,15 @@ func (c *Client) Download(id string, dst io.Writer) error {
 }
 
 // Create creates a new cluster bundle and returns its ID.
-func (c *Client) Create() (string, error) {
-	req, err := c.http.NewRequest("PUT", fmt.Sprintf("%s/%s", baseURL, uuid.NewV4().String()), nil)
+func (c *Client) Create(options Options) (string, error) {
+	opts, err := json.Marshal(options)
+	if err != nil {
+		return "", fmt.Errorf("could not marashal options: %s", opts)
```
makkes:
```suggestion
		return "", fmt.Errorf("could not marshal options: %s", opts)
```

makkes (changes requested):

bamarni:
> Why you haven't raised you objections when we designed API and implemented it?

I've just noticed it when seeing the reason for the previous `--agents=false` implementation. For the API it was just a suggestion, it's fine with me if instead we add more logic into the CLI to improve the UX.

pkg/cmd/diagnostics/diagnostics_create.go
```diff
@@ -26,6 +36,7 @@ func newDiagnosticsCreateCommand(ctx api.Context) *cobra.Command {
 			return err
 		},
 	}
+	cmd.Flags().BoolVar(&masters, "masters", true, "Collect data from masters only")
```
bamarni: This should be false by default now.

pkg/cmd/diagnostics/diagnostics_create.go
janisz: how about `opts.Agents = !masters`

pkg/cmd/diagnostics/diagnostics_create.go
janisz: Fixed

janisz:

pkg/cmd/diagnostics/diagnostics_create.go
```diff
@@ -10,14 +10,18 @@ import (
...
 		Use:   "create",
 		Short: "Create a diagnostics bundle",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			c := pluginutil.HTTPClient("")
 			client := diagnostics.NewClient(c)
-
-			id, err := client.Create()
+			opts := diagnostics.Options{
+				Masters: true,
+				Agents:  !mastersOnly,
```
makkes: Let's think about a possible next step of having `--agents` as well. How would that work? I'd suggest this:

`--masters`: only master nodes
`--agents`: only agent nodes
`--masters --agents`: masters and agents
none: same as `--masters --agents`

Then the code we'd have here would be:

```
opts := diagnostics.Options{
    Masters: mastersOnly || !mastersOnly && !agentsOnly
    Agents: agentsOnly || !mastersOnly && !agentsOnly
}
```
In that case naming the flags `masters` and `agents` instead of `...Only` probably makes more sense.

wdyt?

pkg/diagnostics/v2/client_test.go
```diff
@@ -198,7 +198,7 @@ func TestCreateHappyPath(t *testing.T) {
 	defer ts.Close()
 
 	c := NewClient(pluginutil.HTTPClient(ts.URL))
-	id, err := c.Create()
+	id, err := c.Create(Options{})
```
makkes: We definitely need tests for the correct parsing of the flags.

pkg/cmd/diagnostics/diagnostics_create.go
janisz: Do we need `--agents` flag? The difference between `--agents` and `--masters --agents` is only in 3/5 nodes

makkes: https://jira.mesosphere.com/browse/DCOS_OSS-5547 actually says we can add `--agents` later which makes me think we should at least make it easy to add it later. AFAIK leaving out masters may drastically reduce bundle size, doesn't it?

pkg/diagnostics/v2/client_test.go
janisz: Tests added

janisz:

pkg/cmd/diagnostics/diagnostics_create.go
janisz: Thats true masters are the slowest and heaviest nodes in terms of bundle generation

janisz:

janisz:
Manual tests on latest dcos/master https://github.com/dcos/dcos/pull/6365 that includes https://github.com/dcos/dcos-diagnostics/pull/159

```bash
# dcos diagnostics create
a01fbd7a-42bf-4182-8acc-7ca1ef838d84
# dcos diagnostics create --masters
b8e99df7-2861-47e1-800b-5470b474bd05
# dcos diagnostics create --agents 
0b56b82c-c315-45e5-bfda-7ddcb9cb31e3
# dcos diagnostics create --agents --masters
5a793676-29a7-4c42-bd72-225bec97aa7b
# dcos diagnostics list                     
                   ID                   STATUS     CREATED       SIZE   
  0b56b82c-c315-45e5-bfda-7ddcb9cb31e3  Done    19 seconds ago  1.7 MB  
  5a793676-29a7-4c42-bd72-225bec97aa7b  Done    14 seconds ago  3.0 MB  
  a01fbd7a-42bf-4182-8acc-7ca1ef838d84  Done    31 seconds ago  2.9 MB  
  b8e99df7-2861-47e1-800b-5470b474bd05  Done    24 seconds ago  1.3 MB

# unzip -p a01fbd7a-42bf-4182-8acc-7ca1ef838d84.zip report.json
{"id":"a01fbd7a-42bf-4182-8acc-7ca1ef838d84","nodes":{"172.17.0.2":{"status":"Done"},"172.17.0.3":{"status":"Done"},"172.17.0.4":{"status":"Done"}}}% 
# unzip -p b8e99df7-2861-47e1-800b-5470b474bd05.zip report.json 
{"id":"b8e99df7-2861-47e1-800b-5470b474bd05","nodes":{"172.17.0.2":{"status":"Done"}}}%  
# unzip -p 0b56b82c-c315-45e5-bfda-7ddcb9cb31e3.zip report.json 
{"id":"0b56b82c-c315-45e5-bfda-7ddcb9cb31e3","nodes":{"172.17.0.3":{"status":"Done"},"172.17.0.4":{"status":"Done"}}}%    
# unzip -p 5a793676-29a7-4c42-bd72-225bec97aa7b.zip report.json 
{"id":"5a793676-29a7-4c42-bd72-225bec97aa7b","nodes":{"172.17.0.2":{"status":"Done"},"172.17.0.3":{"status":"Done"},"172.17.0.4":{"status":"Done"}}}% 
```

janisz:

makkes (approved):

janisz:
:wave: @drewvanstone @bamarni @rgo3  Can you review, please?

pkg/diagnostics/v2/client_test.go
```diff
@@ -185,23 +185,43 @@ func TestDeleteWithDeletedBundle(t *testing.T) {
...
+		{expected: `{"masters":true,"agents":true}`, given: Options{Masters: true, Agents: true}},
+	} {
+		t.Run(tc.expected, func(t *testing.T) {
+			ts := httptest.NewServer(handler(t, tc.expected))
+			defer ts.Close()
+			c := NewClient(pluginutil.HTTPClient(ts.URL))
+			id, err := c.Create(tc.given)
+			assert.NoError(t, err)
+			assert.Equal(t, newID.String(), id)
+		})
```
drewvanstone, line 223: Like the table driven tests 👍 !

drewvanstone (approved):
Awesome

PR #372, opened 2019-10-10
Allow bundle creation only for nodes or masters

janisz:
# Backports: #371

* Allow bundle creation for master/agents only

    Previous API allow fine grained control over which nodes data should be included in the bundle. This change adds the most used function to create master only bundle with `dcos diagnostics create --agents=false`.

* Disable gosec on ssh command exec

    DC/OS Cli runs in user context and grabs arguments from user input. We need to trust that users knows what they are doing.


janisz:

drewvanstone (approved):

makkes (approved):

PR #380, opened 2019-10-16
[WIP] Windows on travis

janisz:


drewvanstone:
Mind adding a brief description to the PR? Were windows builds running somewhere else before, and we're migrating to Travis? Or were we not running any builds on windows? 

Also looks like the build is failing

PR #384, opened 2019-10-24
Add more context to create diagnosticsts error.

janisz:
Fixes: https://jira.mesosphere.com/browse/COPS-5439
Refs: https://github.com/dcos/dcos-diagnostics/pull/162


janisz:

pkg/diagnostics/client_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"testing"
+)
+
+func TestClient_CreateErrorsWhenGot503WithStatus(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "/system/health/v1/report/diagnostics/create", r.URL.String())
+		assert.Equal(t, "POST", r.Method)
+		w.WriteHeader(http.StatusServiceUnavailable)
+		_, err := w.Write([]byte(`{
+			"response_http_code":503,
```
makkes, line 18: nit: Looking at the description of https://github.com/dcos/dcos-diagnostics/pull/162 I think that's a 500 now, isn't it?

janisz: It's just for test with different values. Before we have some logic about this codes which was wrong (we were assuming 503 means user provided not existing node),

pkg/diagnostics/client_test.go
```diff
@@ -0,0 +1,78 @@
+package diagnostics
+
+import (
+	"github.com/dcos/dcos-core-cli/pkg/pluginutil"
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"net/http/httptest"
+	"testing"
```
bamarni, line 6: We usually make distinct import groups, but it doesn't seem to be covered by linting currently.

See https://github.com/golang/go/wiki/CodeReviewComments#imports

janisz: Linter is disabled for tests? Should I turn it on?

bamarni: It'd be nice to run some basic linting on tests, maybe not the whole checks that we have, eg. the min-complexity might cause issues on some big table-driven tests.

janisz: Done

pkg/cmd/pkg/package_describe_test.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TesPkgDescribeShouldProxyCommandsToCosmos(t *testing.T) {
+func TestPkgDescribeShouldProxyCommandsToCosmos(t *testing.T) {
```
janisz, line 19: This is why we should run liners on `_test.go`. This typo prevents this test from run. So we will never see it failing.

pkg/diagnostics/client_test.go
makkes: Esp. not `errcheck`.

makkes: Ok, `errcheck` isn't enabled, anyway. 👍🏻

pkg/cmd/pkg/package_list_test.go
```diff
@@ -58,6 +58,59 @@ func TestListPackages(t *testing.T) {
 `, out.String())
 }
 
+var testCases = []struct {
+	cluster string
+	options listOptions
+	out     string
+}{
+	{NoPlugins, listOptions{jsonOutput: true},
+		`[{"apps":["a","b","some-app"],"description":"package-2","framework":false,"name":"package-1","selected":false,"version":"0.0.0.1"},
```
drewvanstone: What do you think about instead of using string literals for the json response, we use cosmos.Package like how it's done on line 29? It'd be clearer what the json is, and it'd be easier to read.

pkg/diagnostics/client_test.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/dcos/dcos-core-cli/pkg/pluginutil"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestClient_CreateErrorsWhenGot503WithStatus(t *testing.T) {
```
drewvanstone: Is there a convention with httptest where you name the test functions with an underscore, e.g. `TestClient_`?

pkg/cmd/pkg/package_list_test.go
janisz: Line 29 is for mocking cosmos resposne. This JSON string is the actual output from CLI. I can change it to a struct but then it will be converted to JSON.

drewvanstone: Ya fair enough. So do you prefer to keep it as a string literal?

janisz: Fixed

pkg/diagnostics/client_test.go
janisz: Fixed

janisz:

pkg/cmd/pkg/package_describe_test.go
makkes: Which linter checks this?

makkes (approved):

drewvanstone (approved):

PR #387, opened 2019-10-29
Generate diagnostics bundle after E2E failure

janisz:
* Use `dcos diagnostics create` after test failure
* Update dcos-launch and dcos-test-utils

Fixes:

* https://jira.mesosphere.com/browse/DCOS_OSS-5599
* https://jira.mesosphere.com/browse/DCOS_OSS-5598


bamarni:
How does it fix the linked JIRA tickets? These dependencies are used to spin-up the cluster and the flaky network errors we might get happen afterwards, when running tests against the cluster.

ci/integration-tests-linux.groovy
```diff
@@ -1,68 +1,118 @@
...
-      agent { label 'mesos' }
+    stages {
+        stage("Build Go binary") {
+            steps {
+                sh 'make linux'
+                stash includes: 'build/linux/**', name: 'dcos-linux'
+            }
+        }
+
+        stage("Lanuch AWS Cluster") {
```
drewvanstone:
```suggestion
        stage("Launch AWS Cluster") {
```

janisz:
I clean up script a little bit. Now pipelines for linux and mac are nearly the same and there are minor differences with windows. 

<details>

```diff
diff -d  ci/integration-tests-linux.groovy ci/integration-tests-mac.groovy  
24c24
< def os = 'linux'
---
> def os = 'darwin'
66c66
<             agent { label 'py36' }
---
>             agent { label 'mac-hh-yosemite' }
```

```
diff3 ci/integration-tests-*              
====
1:24c
  def os = 'linux'
2:24c
  def os = 'darwin'
3:24c
  def os = 'windows'
====3
1:64a
2:64a
3:65,71c
          stage("Stash dcos.exe") {
              steps {
                  sh "wget -q https://downloads.dcos.io/cli/testing/binaries/dcos/${os}/x86-64/master/dcos.exe"
                  stash includes: 'dcos.exe', name: 'dcos-exe'
              }
          }
  
====
1:66c
              agent { label 'py36' }
2:66c
              agent { label 'mac-hh-yosemite' }
3:73c
              agent { label 'windows' }
====3
1:67a
2:67a
3:75c
                  unstash "dcos-exe"
====3
1:71c
2:71c
                          sh '''scripts/run_integration_tests.sh'''
3:79,95c
                          bat '''
                              bash -exc " \
                              export PYTHONIOENCODING=utf-8; \
                              export CLI_TEST_SSH_USER=centos; \
                              export CLI_TEST_MASTER_PROXY=1; \
                              mkdir -p build/windows; \
                              make python; \
                              python scripts/plugin/package_plugin.py; \
                              cd python/lib/dcoscli; \
                              make env; \
                              rm -f ./env/Scripts/dcos.exe; \
                              mv ../../../dcos.exe dist; \
                              PATH=$PWD/dist:$PATH; \
                              dcos cluster remove --all; \
                              dcos cluster setup --no-check ${DCOS_TEST_URL}; \
                              dcos plugin add -u ../../../build/$OS/dcos-core-cli.zip; \
                              ./env/Scripts/pytest -vv -x --durations=10 -p no:cacheprovider tests/integrations"'''
```

</details>

janisz:
:wave: @drewvanstone @br-lewis @makkes @bamarni @rgo3  @ArmandGrillet Can you review, please?

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+export PYTHONIOENCODING=utf-8
+export CLI_TEST_SSH_USER=centos
+export CLI_TEST_MASTER_PROXY=1
+DCOS_EXPERIMENTAL=1 make plugin
```
bamarni, line 6: Why do we set `DCOS_EXPERIMENTAL`? This means it doesn't verify that the Python version is 3.7, and we might run tests with a CLI having a Python version different from the one we use for building releases.

ci/integration-tests-linux.groovy
```diff
@@ -1,68 +1,114 @@
...
-    timeout(time: 6, unit: 'HOURS')
-  }
+def master_ip = 'UNKNOWN'
+def os = 'linux'
 
-  stages {
-    stage("Build Go binary") {
-      agent { label 'mesos-ubuntu' }
+pipeline {
+    agent { label 'mesos' }
```
bamarni, line 27: Previously we were building the binary from `mesos-ubuntu`, why switching to `mesos`? IIRC `mesos` relies on an alpine container, and building from there was creating binaries which wouldn't run on unbuntu.

We should run integration tests with a binary built the same way than the one we release.

makkes (approved):

ci/integration-tests-windows.groovy
```diff
@@ -1,107 +1,138 @@
...
+            steps {
+                sh "make ${os}"
+                stash includes: "build/${os}/**", name: "dcos-${os}"
+            }
+        }
 
-                        // Cannot `checkout scm` here.
-                        // https://mesosphere.slack.com/archives/C5U03P5T6/p1527867956000237
-                        unstash "dcos-core-cli"
+        stage("Lanuch AWS Cluster") {
```
armandgrillet: Launch*

scripts/run_integration_tests.sh
```diff
@@ -0,0 +1,14 @@
...
+cd python/lib/dcoscli
+source env/bin/activate
+wget -qO env/bin/dcos https://downloads.dcos.io/cli/testing/binaries/dcos/${OS}/x86-64/master/dcos
+chmod +x env/bin/dcos
+dcos cluster remove --all
+dcos cluster setup --no-check ${DCOS_TEST_URL}
+dcos plugin add -u ../../../build/$OS/dcos-core-cli.zip
+py.test -vv -x --durations=10 -p no:cacheprovider tests/integrations
```
armandgrillet: Missing jump line at the end of the file.

armandgrillet (approved):
LGTM

ci/integration-tests-windows.groovy
janisz: Fixed, it was wrong in all 3 files

scripts/run_integration_tests.sh
janisz: Added new line

ci/integration-tests-linux.groovy
```diff
@@ -1,68 +1,114 @@
...
-                dcos cluster remove --all; \
-                ./run_integration_tests.py"
-          '''
+        stage("Run integration tests") {
+            agent { label 'py36' }
+            steps {
+                unstash "dcos-${os}"
+                withEnv(["DCOS_TEST_URL=${master_ip}", "OS=${os}"]) {
+                    withCredentials(credentials) {
+                        sh '''scripts/run_integration_tests.sh'''
```
bamarni, line 71: Regarding moving scripts from Python to Bash, I think in general it creates a bigger gap between UNIX and Windows Jenkinsfile / pipelines, as it makes it hard to share the same script. Even though we have winbash installed on windows nodes, Python is much more easier to run cross-platforms.

ci/integration-tests-linux.groovy
bamarni: I think it was because the binary was dynamically linked and the glibc / musl differences.

scripts/run_integration_tests.sh
janisz: By adding `DCOS_EXPERIMENTAL=1` we can have the same script for mac and linux.

ci/integration-tests-linux.groovy
janisz: I switched to `mesos` from `mesos-ubuntu` since we don't need the whole ubuntu software to build binary and can rely on lightweight system.

ci/integration-tests-linux.groovy
janisz: Good point but previously we have `run_integration_tests` in Python and Windows was not using it. We can think about moving from bash to python to support windows but I'm not sure if that's cross platform since

1.  dcos binary has `.exe` extension on Windows
2. On Mac/Linux we use `make plugin` while on Windows we are using `make python`

Full diff:

```diff
0a1,2
> #!/usr/bin/env bash
> 
4,6c6
< mkdir -p build/windows
< make python
< python scripts/plugin/package_plugin.py
---
> DCOS_EXPERIMENTAL=1 make plugin
8,11c8,10
< make env
< rm -f ./env/Scripts/dcos.exe
< mv ../../../dcos.exe dist
< PATH=$PWD/dist:$PATH
---
> source env/bin/activate
> wget -qO env/bin/dcos https://downloads.dcos.io/cli/testing/binaries/dcos/${OS}/x86-64/master/dcos
> chmod +x env/bin/dcos
15,16c14
< ./env/Scripts/pytest -vv -x --durations=10 -p no:cacheprovider tests/integrations
< 
---
> py.test -vv -x --durations=10 -p no:cacheprovider tests/integrations
```

ci/integration-tests-linux.groovy
janisz: Ideal situation will be to use released version for integration tests. In fact we can have everything in a single pipeline. Right now we have binary build stage in 4 places, and E2E tests for different platforms shares 99% of code.

ci/integration-tests-linux.groovy
bamarni: closing as it's more of a general comment, indeed currently they are not fully shared yet

ci/integration-tests-linux.groovy
bamarni: As discussed, even though we use an alpine node it eventually runs in an Ubuntu container, so that's similar to the released one.

bamarni (approved):
As discussed, with these changes we wouldn't build the plugin used for tests with the same Python version as the one in released plugins (Python 3.7), which is not ideal but should be fixed eventually once the marathon subcommand is refactored in Go (as we wouldn't build Python code anymore).

makkes:
> As discussed, with these changes we wouldn't build the plugin used for tests with the same Python version as the one in released plugins (Python 3.7), which is not ideal but should be fixed eventually once the marathon subcommand is refactored in Go (as we wouldn't build Python code anymore).

As far as I  know the Windows binary is built with Python 3.5 (https://github.com/dcos/dcos-core-cli/blob/4f7337425d6506bbbf8f7d31b89056307e344d4d/python/bin/env.sh#L26), anyway so we have deviations already.

PR #388, opened 2019-10-29
[WIP] Use terraform to run CI

janisz:


scripts/launch_aws_cluster.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set +x
+
+bash <(curl -s https://raw.githubusercontent.com/dcos/dcos/master/test_util/terraform_init.sh)
+
+./terraform init
```
makkes: If I understand this line correctly, this is already done here:

https://github.com/dcos/dcos/blob/09608a1c0e0bf2237ffd0afb1111faa3414a6abb/test_util/terraform_init.sh#L79

scripts/launch_aws_cluster.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set +x
+
+bash <(curl -s https://raw.githubusercontent.com/dcos/dcos/master/test_util/terraform_init.sh)
+
+./terraform apply
```
makkes:
```suggestion
./terraform apply --auto-approve
```

makkes (changes requested):

scripts/launch_aws_cluster.sh
```diff
@@ -0,0 +1,70 @@
...
+URL_terraform_provider_aws="https://downloads.mesosphere.com/dcos-terraform/terraform/v0.11.13-mesosphere/${OS}_amd64/terraform-provider-aws"
+
+if [ "$OS" == "linux" ]; then
+    CHECKSUM_terraform="19b19b9c979a75b523ce8ad044bade7a0987ce38f866cfe5479068531839f70c"
+    CHECKSUM_terraform_provider_aws="245ba160a0389c2177091e2dbc72d79d034f1b0a93a5fee01d0d4b69f22be4e1"
+elif [ "$OS" == "darwin" ]; then
+    CHECKSUM_terraform="9bf9567bb0d396f83dd8ee2cedde5f188bdbba3316528648798e1d3f6814a1f7"
+    CHECKSUM_terraform_provider_aws="bdb07595b393c2fba7e0a679e1d9355e9b30edeec138f48fd1ff9deb2065a8f9"
+else
+    # How did you even get here?
```
drewvanstone, line 26: 😄

bamarni: What we are currently doing for Windows is spin-up the cluster from a Linux machine and then storing the master node IP in a variable which is read from the Windows node:

https://github.com/dcos/dcos-core-cli/blob/2bc9f1b3f61fe50a4c863ae2f176c30815d473b8/ci/integration-tests-windows.groovy#L47-L56

This is not ideal but was done like this because dcos-launch doesn't run on Windows, it for example forced us to use the scripted pipeline instead of the declarative pipeline like other platforms.

On Windows, if we can spin-up the cluster from the node running the tests directly without having this multi-node steps, I think that's preferable. At least we seem to publish a binary for Windows as well (https://downloads.mesosphere.com/dcos-terraform/terraform/v0.11.13-mesosphere/windows_amd64/terraform.exe).

janisz: Take a look at https://github.com/dcos/dcos-core-cli/pull/387 it's the first step to unify windows Mac and Linux integration tests.

PR #393, opened 2019-11-06
Port 'dcos marathon about' in Go.

armandgrillet:
https://jira.mesosphere.com/browse/DCOS-60268

### DIFF
--- a/pkg/cmd/marathon/marathon_about.go
+++ b/pkg/cmd/marathon/marathon_about.go
@@ -1,8 +1,11 @@
 package marathon
 
 import (
+	"encoding/json"
+
 	"github.com/dcos/dcos-cli/api"
-	"github.com/dcos/dcos-core-cli/pkg/cmd/marathon/python"
+	"github.com/dcos/dcos-core-cli/pkg/marathon"
+
 	"github.com/spf13/cobra"
 )
 
@@ -11,7 +14,23 @@ func newCmdMarathonAbout(ctx api.Context) *cobra.Command {
 		Use:   "about",
 		Short: "Print info.json for DC/OS Marathon.",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return python.InvokePythonCLI(ctx)
+			client, err := marathon.NewClient(ctx)
+			if err != nil {
+				return err
+			}
+
+			return marathonAbout(ctx, *client)
 		},
 	}
+}
+
+func marathonAbout(ctx api.Context, client marathon.Client) error {
+	info, err := client.Info()
+	if err != nil {
+		return err
+	}
+
+	enc := json.NewEncoder(ctx.Out())
+	enc.SetIndent("", "    ")
+	return enc.Encode(info)
 }
--- a/pkg/cmd/marathon/marathon_about_test.go
+++ b/pkg/cmd/marathon/marathon_about_test.go
@@ -1,0 +1,55 @@
+package marathon
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/dcos/dcos-cli/pkg/config"
+	"github.com/dcos/dcos-cli/pkg/mock"
+	"github.com/dcos/dcos-core-cli/pkg/marathon"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMarathonAbout(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "/service/marathon/v2/info", r.URL.String())
+		assert.Equal(t, "GET", r.Method)
+		w.Write([]byte(`{"frameworkId":"marathonId"}`))
+	}))
+	defer ts.Close()
+
+	env := mock.NewEnvironment()
+	ctx := mock.NewContext(env)
+	cluster := config.NewCluster(nil)
+	cluster.SetURL(ts.URL)
+	ctx.SetCluster(cluster)
+
+	client, err := marathon.NewClient(ctx)
+	assert.NoError(t, err)
+
+	err = marathonAbout(ctx, *client)
+	assert.NoError(t, err)
+}
+
+func TestMarathonAboutClientError(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "/service/marathon/v2/info", r.URL.String())
+		assert.Equal(t, "GET", r.Method)
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte("500 - Something bad happened!"))
+	}))
+	defer ts.Close()
+
+	env := mock.NewEnvironment()
+	ctx := mock.NewContext(env)
+	cluster := config.NewCluster(nil)
+	cluster.SetURL(ts.URL)
+	ctx.SetCluster(cluster)
+
+	client, err := marathon.NewClient(ctx)
+	assert.NoError(t, err)
+
+	err = marathonAbout(ctx, *client)
+	assert.EqualError(t, err, `HTTP 500 error`)
+}
--- a/pkg/marathon/client.go
+++ b/pkg/marathon/client.go
@@ -3,9 +3,13 @@ package marathon
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
 	"strings"
 
 	"github.com/dcos/dcos-cli/api"
+	"github.com/dcos/dcos-cli/pkg/httpclient"
 	"github.com/dcos/dcos-core-cli/pkg/pluginutil"
 	marathon "github.com/gambol99/go-marathon"
 )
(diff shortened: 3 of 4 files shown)


pkg/cmd/marathon/marathon_about.go
```diff
@@ -11,7 +14,19 @@ func newCmdMarathonAbout(ctx api.Context) *cobra.Command {
 		Use:   "about",
 		Short: "Print info.json for DC/OS Marathon.",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return python.InvokePythonCLI(ctx)
+			client, err := marathon.NewClient(ctx)
```
makkes, line 17: I'd prefer to have this function consist of a single function call just like in https://github.com/dcos/dcos-core-cli/blob/2bc9f1b3f61fe50a4c863ae2f176c30815d473b8/pkg/cmd/pkg/package_install.go#L38 and then let's test this functionality as well, please.

makkes (changes requested):

pkg/marathon/client.go
```diff
@@ -64,5 +65,31 @@ func (c *Client) GroupsWithoutRootSlash() (map[string]bool, error) {
...
 		return nil, errors.New("unable to get Marathon groups")
 	}
+}
+
+// Info returns the content of the Marathon endpoint 'v2/info'.
+func (c *Client) Info() (map[string]interface{}, error) {
+	dcosClient := pluginutil.HTTPClient(c.baseURL)
+	resp, err := dcosClient.Get("/v2/info")
+	if err != nil {
+		return nil, err
```
janisz, line 78: Please add more context to this error

pkg/marathon/client.go
```diff
@@ -64,5 +65,31 @@ func (c *Client) GroupsWithoutRootSlash() (map[string]bool, error) {
...
+		}
 
+		var result map[string]interface{}
+		err = json.Unmarshal(data, &result)
+		if err != nil {
+			return nil, err
+		}
+		return result, nil
+	default:
+		return nil, errors.New("unable to get Marathon info")
```
janisz: How about adding first bytes of the response or at least status code. It's hard to tell what went wrong if you get `unable to get Marathon info` error. See: https://jira.mesosphere.com/browse/COPS-5439

janisz (changes requested):

pkg/marathon/client.go
armandgrillet: The error reported by the HTTPClient should give enough information and this is how GET calls are handled in the rest of the codebase (e.g. https://github.com/dcos/dcos-core-cli/blob/2bc9f1b3f61fe50a4c863ae2f176c30815d473b8/pkg/diagnostics/v2/client.go#L53). Could you tell me what kind of information should I add?

janisz: OK let's leave it for now. This is a prat of bigger discussion how to handle errors in CLI

pkg/marathon/client.go
```diff
@@ -64,5 +68,40 @@ func (c *Client) GroupsWithoutRootSlash() (map[string]bool, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		return result, nil
+	default:
+		return nil, httpResponseToError(resp)
+	}
+}
 
+func httpResponseToError(resp *http.Response) error {
```
janisz, line 100: Please add a test for this

pkg/marathon/client.go
```diff
@@ -64,5 +68,40 @@ func (c *Client) GroupsWithoutRootSlash() (map[string]bool, error) {
...
+
+	switch resp.StatusCode {
+	case 200:
+		data, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, err
+		}
+
+		var result map[string]interface{}
+		err = json.Unmarshal(data, &result)
```
janisz, line 90: What if marathon returns not a json? Will the error be clear for user?

pkg/cmd/marathon/marathon_about.go
```diff
@@ -11,7 +14,23 @@ func newCmdMarathonAbout(ctx api.Context) *cobra.Command {
...
+
+			return marathonAbout(ctx, *client)
 		},
 	}
 }
+
+func marathonAbout(ctx api.Context, client marathon.Client) error {
+	info, err := client.Info()
+	if err != nil {
+		return err
```
janisz, line 30: Add a test for this

pkg/cmd/marathon/marathon_about_test.go
```diff
@@ -0,0 +1,50 @@
...
+	}))
+	defer ts.Close()
+
+	var out bytes.Buffer
+	env := mock.NewEnvironment()
+	env.Fs = afero.NewCopyOnWriteFs(
+		afero.NewReadOnlyFs(afero.NewOsFs()),
+		afero.NewMemMapFs(),
+	)
+	env.Out = &out
```
janisz:
```diff
diff --git a/pkg/cmd/marathon/marathon_about_test.go b/pkg/cmd/marathon/marathon_about_test.go
index 4b424aa0..1a762cba 100644
--- a/pkg/cmd/marathon/marathon_about_test.go
+++ b/pkg/cmd/marathon/marathon_about_test.go
@@ -1,7 +1,6 @@
 package marathon
 
 import (
-       "bytes"
        "encoding/json"
        "net/http"
        "net/http/httptest"
@@ -10,7 +9,6 @@ import (
        "github.com/dcos/dcos-cli/pkg/config"
        "github.com/dcos/dcos-cli/pkg/mock"
        "github.com/dcos/dcos-core-cli/pkg/marathon"
-       "github.com/spf13/afero"
        "github.com/stretchr/testify/assert"
 )
 
@@ -29,13 +27,7 @@ func TestMarathonAbout(t *testing.T) {
        }))
        defer ts.Close()
 
-       var out bytes.Buffer
        env := mock.NewEnvironment()
-       env.Fs = afero.NewCopyOnWriteFs(
-               afero.NewReadOnlyFs(afero.NewOsFs()),
-               afero.NewMemMapFs(),
-       )
-       env.Out = &out
 
        ctx := mock.NewContext(env)
        cluster := config.NewCluster(nil)
```

pkg/cmd/marathon/marathon_about_test.go
```diff
@@ -0,0 +1,50 @@
...
+		FrameworkID string `json:"frameworkId"`
+	}
+	marathonInfo := info{
+		FrameworkID: "marathonId",
+	}
+
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "/service/marathon/v2/info", r.URL.String())
+		assert.Equal(t, "GET", r.Method)
+		assert.NoError(t, json.NewEncoder(w).Encode(&marathonInfo))
```
janisz:
```suggestion
		w.Write([]byte(`{"frameworkId":"marathonId"}`))
```

janisz: We don't need this structures since we are not testing if golang json marshal is working but if real response from marathon will be parsed properly.

pkg/marathon/client.go
```diff
@@ -64,5 +68,40 @@ func (c *Client) GroupsWithoutRootSlash() (map[string]bool, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	switch resp.StatusCode {
+	case 200:
+		data, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, err
```
janisz:
```suggestion
			return nil, fmt.Errorf("could not read resposne body: %s", err)
```

pkg/marathon/client.go
```diff
@@ -64,5 +68,40 @@ func (c *Client) GroupsWithoutRootSlash() (map[string]bool, error) {
...
+		}
+
+		var result map[string]interface{}
+		err = json.Unmarshal(data, &result)
+		if err != nil {
+			return nil, err
```
janisz:
```suggestion
			return nil, fmt.Errorf("could not unmarshal resposne body: %s", err)
```

pkg/marathon/client.go
```diff
@@ -64,5 +68,40 @@ func (c *Client) GroupsWithoutRootSlash() (map[string]bool, error) {
...
+	default:
+		return nil, httpResponseToError(resp)
+	}
+}
 
+func httpResponseToError(resp *http.Response) error {
+	if resp.StatusCode < 400 {
+		return fmt.Errorf("unexpected status code %d", resp.StatusCode)
+	}
+	return &httpclient.HTTPError{
```
janisz, line 104: I really don't like this error it has no more information then one above

https://github.com/dcos/dcos-core-cli/blob/55b3085ed15a1b26df8f0fbce58c15958c2541cc/vendor/github.com/dcos/dcos-cli/pkg/httpclient/client.go#L55-L57

janisz: But this is the part of bigger discussion how to handle API errors

janisz: In marathon I think we should stream resposne to a client. E.g. 500 error can happen when there is no leader and this will be clear to the user whats wrong

armandgrillet: IMO we should do like in the Mesos client (e.g. https://github.com/dcos/dcos-core-cli/blob/4f7337425d6506bbbf8f7d31b89056307e344d4d/pkg/mesos/client.go#L134), if we see specific HTTP code in the Marathon REST API defined in https://mesosphere.github.io/marathon/api-console/index.html we should handle them in the client. 

In the case of `v2/info`, only `200` is expected which explains the lack of further handling.

makkes (approved):

janisz (approved):

PR #389, opened 2019-10-30
Backport bugfix to `dcos package list` to 2.0

bamarni:
This backports a bugfix for the `dcos package list` command, where the
`--app-id` option was not working.

As this fix was part of a more general refactoring PR that got
squashed (#378), and I've simply cherry-picked the commit, this
also introduces a new feature. I've added it in the CHANGELOG.

https://jira.mesosphere.com/browse/COPS-5502

PR #396, opened 2019-11-07
Make local installed apps override Cosmos Apps

janisz:
`dcos package --json` and `dcos package --cli --json` may results in different resposnes. To keep backward compatibility we need to make locally installed packages overrides cosmos response.
This happens only for packages that has cli and app.

Fixes: [DCOS_OSS-5657](https://jira.mesosphere.com/browse/DCOS_OSS-5657)
Reverts: https://github.com/dcos/dcos-core-cli/pull/389/commits/5cb304b532da5b22b40d21bcb9716c60bf24f9e1


bamarni (approved):
Nice one :+1: 

Should also be backported to 2.0-patch.x

makkes (approved):

PR #402, opened 2019-11-13
Create 'marathon app add'

makkes:
closes https://jira.mesosphere.com/browse/DCOS-60269


pkg/cmd/marathon/app/app_add.go
```diff
@@ -11,8 +12,19 @@ func newCmdMarathonAppAdd(ctx api.Context) *cobra.Command {
 		Use:   "add",
 		Short: "Add an application.",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return python.InvokePythonCLI(ctx)
+			client, err := marathon.NewClient(ctx)
+			if err != nil {
+				return err
+			}
+			if len(args) > 0 {
```
armandgrillet, line 23: I would use `Args:  cobra.ExactArgs(1),` (like in `job_remove.go`) instead of having this check to use the default cobra args handling.

makkes: I don't know how that would work since the argument can just be empty and then the JSON is read from stdin.

br-lewis: There's also a maximum number of args function. `cobra.MaximumNArgs(n)` maybe?

.gitignore
```diff
@@ -18,3 +18,4 @@ bundle.zip
 Vagrantfile
 .vagrant
 .idea
+.vscode
```
janisz: nit: https://help.github.com/en/github/using-git/ignoring-files#create-a-global-gitignore

pkg/marathon/client.go
```diff
@@ -43,6 +45,81 @@ func NewClient(ctx api.Context) (*Client, error) {
 	return &Client{API: client, baseURL: baseURL}, nil
 }
 
+func fetchAppDefinition(url string) ([]byte, error) {
```
janisz: it's easier to read if helper function is below it's usage

pkg/marathon/client.go
```diff
@@ -43,6 +45,81 @@ func NewClient(ctx api.Context) (*Client, error) {
...
+	resp, err := http.Get(path.Clean(url))
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("Could not fetch app definition from %s", url)
```
janisz:
```suggestion
		return nil, fmt.Errorf("could not fetch app definition from %s", url)
```

pkg/marathon/client.go
```diff
@@ -43,6 +45,81 @@ func NewClient(ctx api.Context) (*Client, error) {
...
+		appBytes, err = ioutil.ReadAll(ctx.Input())
+		if err != nil {
+			return nil, err
+		}
+		return appBytes, nil
+	}
+
+	// get definition from file and fall back to HTTP
+	appBytes, err = ioutil.ReadFile(location)
+	if err != nil {
```
janisz: No test for this case

pkg/marathon/client.go
```diff
@@ -43,6 +45,81 @@ func NewClient(ctx api.Context) (*Client, error) {
...
+	if location == "" {
+		appBytes, err = ioutil.ReadAll(ctx.Input())
+		if err != nil {
+			return nil, err
+		}
+		return appBytes, nil
+	}
+
+	// get definition from file and fall back to HTTP
+	appBytes, err = ioutil.ReadFile(location)
```
janisz:
```go
	appBytes, err = ioutil.ReadFile(location)
	if err == nil {
		return appBytes, nil
	}
	if !os.IsNotExist(err) {
		return nil, err
	}
	return fetchAppDefinition(location)
```

janisz: We can even inline `fetchAppDefinition`

pkg/marathon/client.go
```diff
@@ -43,6 +44,81 @@ func NewClient(ctx api.Context) (*Client, error) {
...
+	var app marathon.Application
+	err = json.Unmarshal(appBytes, &app)
+	if err != nil {
+		return err
+	}
+
+	existingApp, err := c.API.ApplicationBy(app.ID, &marathon.GetAppOpts{})
+	if err != nil {
+		if apiErr, ok := err.(*marathon.APIError); ok {
+			if apiErr.ErrCode != marathon.ErrCodeNotFound {
```
janisz: I don't get it. Can you add a test for this case?

pkg/marathon/client.go
```diff
@@ -43,6 +44,81 @@ func NewClient(ctx api.Context) (*Client, error) {
...
+				return err
+			}
+		} else {
+			return err
+		}
```
janisz, line 114: How about simplifying it to
```go 
		if apiErr, ok := err.(*marathon.APIError); !ok {
			return err
		} else if apiErr.ErrCode != marathon.ErrCodeNotFound {
			return err
		}
```

pkg/marathon/client.go
```diff
@@ -43,6 +44,81 @@ func NewClient(ctx api.Context) (*Client, error) {
...
+	if existingApp != nil {
+		return fmt.Errorf("application '%s' already exists", app.ID)
+	}
+
+	createdApp, err := c.API.CreateApplication(&app)
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(ctx.Out(), "Created deployment %s\n", createdApp.Deployments[0]["id"])
```
janisz: I think the client should not interact with user. This should be printed in command.
```suggestion
```

pkg/mocks/context.go
```diff
@@ -0,0 +1,144 @@
+package mocks
```
janisz: Do we need this? See:https://github.com/dcos/dcos-core-cli/blob/55b3085ed15a1b26df8f0fbce58c15958c2541cc/vendor/github.com/dcos/dcos-cli/pkg/mock/mock.go#L82-L92

janisz (changes requested):

pkg/marathon/client.go
makkes: There is a test for both cases:
* TestAddAppWithAppFetchingError (an error which is not a 404)
* TestAddAppWithAppNotFoundError (a 404 effectively)

janisz: Yeah but no test for `apiErr.ErrCode != marathon.ErrCodeNotFound` please add one. You can parametrize `TestAddAppWithAppFetchingError` and use `errors.New(...` and `marathon.ErrCodeServer`

pkg/marathon/client_test.go
```diff
@@ -1,9 +1,17 @@
...
 
 import (
+	"fmt"
+	"io"
 	"net/http"
+	"strings"
 	"testing"
 
+	"github.com/dcos/dcos-core-cli/pkg/mocks"
+	marathon "github.com/gambol99/go-marathon"
```
armandgrillet: We should come up with a name that we use everywhere when importing `github.com/gambol99/go-marathon` as having it as `marathon` here while being in a package called `marathon` makes the situation confusing. I'm thinking about "goMarathon`, what do you think? cc @janisz

pkg/marathon/client.go
```diff
@@ -6,43 +6,112 @@ import (
...
+}
+
 // NewClient creates a new HTTP wrapper client to talk to the Marathon service.
 func NewClient(ctx api.Context) (*Client, error) {
 	cluster, err := ctx.Cluster()
 	if err != nil {
 		return nil, err
 	}
-	baseURL, _ := cluster.Config().Get("marathon.url").(string)
+	baseURL, _ := cluster.Config().Get("goMarathon.url").(string)
```
janisz: Are you sure?

pkg/marathon/client_test.go
```diff
@@ -16,3 +23,73 @@ func TestResponseToError(t *testing.T) {
...
+		{
+			name:             "application does not exists, yet",
+			applicationByErr: &goMarathon.APIError{ErrCode: goMarathon.ErrCodeNotFound},
+			applicationByApp: nil,
+			expectedApp:      &goMarathon.Application{ID: "some id"},
+			expectedErr:      nil,
+		},
+		{
+			name:             "arbitrary APIError",
+			applicationByErr: &goMarathon.APIError{ErrCode: 999},
```
janisz, line 325: Please add one more test wiht error that is not `goMarathon.APIError` to check all cases

janisz (commented):
Only 2 small comments

pkg/marathon/client.go
makkes: I search/replaced too blindly

janisz (approved):

pkg/marathon/client.go
```diff
@@ -31,18 +49,87 @@ func NewClient(ctx api.Context) (*Client, error) {
...
+	switch locationURL.Scheme {
+	case "":
+		appBytes, err = ioutil.ReadFile(location)
+		if err == nil {
+			return appBytes, nil
+		}
+		if !os.IsNotExist(err) {
+			return nil, err
+		}
+	case "http", "https":
```
janisz: Does the python version require scheme?

makkes: Implicitly, yes. It only supports HTTP: https://github.com/dcos/dcos-core-cli/blob/86fe4f9c213d597adbc8daf9c0a672c871b24711/python/lib/dcoscli/dcoscli/marathon/main.py#L306

br-lewis (approved):

makkes:
I finally figured out why Windows CI checks are failing. Fixing those now.

pkg/cmd/marathon/app/app_add.go
```diff
@@ -1,18 +1,52 @@
...
+				fmt.Fprintf(ctx.ErrOut(), "%s", err)
+				return
+			}
+			if len(args) > 0 {
+				err = marathonAppAdd(ctx, *client, args[0])
+			} else {
+				err = marathonAppAdd(ctx, *client, "")
+			}
+			if err != nil {
+				os.Exit(1)
```
janisz: Exithing without any error message? This might be confusing. Why not using `RunE` like it's done in other commands?

pkg/cmd/marathon/app/app_add.go
```diff
@@ -1,18 +1,52 @@
...
+
+func marathonAppAdd(ctx api.Context, client marathon.Client, appFile string) error {
+	app, err := client.AddApp(ctx, appFile)
+	if err != nil {
+		if err == marathon.ErrCannotReadAppDefinition {
+			fmt.Fprintf(ctx.ErrOut(), "Can't read from resource: %s.\nPlease check that it exists.\n", appFile)
+		} else if syntaxError, ok := err.(*json.SyntaxError); ok {
+			fmt.Fprintf(ctx.ErrOut(), "Error loading JSON: %s\n", syntaxError.Error())
+		} else {
+			fmt.Fprintf(ctx.ErrOut(), "%s\n", err.Error())
```
janisz: Look at my comment above. All this `fmt.Printf` should be changed to `return fmt.Errorf`
```suggestion
			return err
```

janisz (changes requested):

pkg/cmd/marathon/app/app_add.go
janisz:
```
func newCmdMarathonAppAdd(ctx api.Context) *cobra.Command {
	cmd := &cobra.Command{
		Use:   "add",
		Short: "Add an application.",
		Args:  cobra.MaximumNArgs(1),
			client, err := marathon.NewClient(ctx)
			if err != nil {
				return err
			}
			if len(args) > 0 {
				err = marathonAppAdd(ctx, *client, args[0])
			} else {
				err = marathonAppAdd(ctx, *client, "")
			}
			return err
		},
	}
	return cmd
}

func marathonAppAdd(ctx api.Context, client marathon.Client, appFile string) error {
	app, err := client.AddApp(ctx, appFile)
	if err != nil {
		if err == marathon.ErrCannotReadAppDefinition {
			return fmt.Errorf("Can't read from resource: %s.\nPlease check that it exists.\n", appFile)
		}
		if syntaxError, ok := err.(*json.SyntaxError); ok {
			return fmt.Errorf("Error loading JSON: %s\n", syntaxError.Error())
		}
		return err
	}
	fmt.Fprintf(ctx.Out(), "Created deployment %s\n", app.Deployments[0]["id"])
	return nil
}
```

pkg/marathon/client.go
```diff
@@ -6,20 +6,35 @@ import (
...
+	"net/url"
+	"os"
+	"regexp"
 	"strings"
 
 	"github.com/dcos/dcos-cli/api"
 	"github.com/dcos/dcos-cli/pkg/httpclient"
 	"github.com/dcos/dcos-core-cli/pkg/pluginutil"
-	"github.com/gambol99/go-marathon"
+	goMarathon "github.com/gambol99/go-marathon"
```
janisz, line 17: Do we need this change?

pkg/cmd/marathon/app/app_add.go
makkes: Oh shit, that was a copy'n'paste mistake.

pkg/marathon/client.go
makkes: We agreed on renaming this import, so yes.

pkg/cmd/marathon/app/app_add.go
janisz: After moving to `RunE` you can easily add tests

makkes: I don't agree on giving up tight control over the messages a user sees. Also, using just `error`s here will force us to format user-visible output according to common error-formatting rules such as not starting an error message with capital letters or ending it with a `.` which doesn't make sense in this scenario.

janisz (approved):

PR #403, opened 2019-11-14
Port `dcos marathon app kill` command to Go

armandgrillet:


janisz (changes requested):
Please add tests

pkg/cmd/marathon/app/app_kill.go
```diff
@@ -13,12 +16,45 @@ func newCmdMarathonAppKill(ctx api.Context) *cobra.Command {
...
+
+				enc := json.NewEncoder(ctx.Out())
+				enc.SetIndent("", "    ")
+				return enc.Encode(deployment)
+			}
+
+			tasks, err := client.KillTasks(args[0], scale, host)
+			b, _ := json.Marshal(tasks)
+			fmt.Printf("Killed tasks: [%s]\n", string(b))
+			return nil
```
br-lewis: The linter caught that you should be returning `err` here.

janisz (approved):

pkg/cmd/marathon/app/app_kill.go
```diff
@@ -13,13 +17,62 @@ func newCmdMarathonAppKill(ctx api.Context) *cobra.Command {
...
+	cmd.Flags().BoolVar(&scale, "scale", false, "Scale the app down after performing the operation.")
 	cmd.Flags().StringVar(&host, "host", "", "The hostname that is running app.")
 
 	return cmd
 }
+
+func marathonAppKill(ctx api.Context, client marathon.Client, appID string, scale bool, host string) error {
+	if scale {
+		_, err := client.API.Application(appID)
+		if apiErr, ok := err.(*goMarathon.APIError); ok && apiErr.ErrCode == goMarathon.ErrCodeNotFound {
```
makkes, line 46: What if there's another error? Shouldn't we exit in those cases as well?

makkes (changes requested):

pkg/cmd/marathon/app/app_kill.go
janisz: :+1: @ArmandGrillet Can you add a test for this case?

makkes (approved):

PR #407, opened 2019-11-19
Change marathon app show into Go

br-lewis:
Addresses https://jira.mesosphere.com/browse/DCOS-60274


pkg/marathon/client.go
```diff
@@ -160,3 +160,55 @@ func (c *Client) Queue() (marathon.Queue, error) {
...
+		var result marathon.ApplicationVersions
+		err = json.NewDecoder(resp.Body).Decode(&result)
+		return result, err
+	default:
+		return marathon.ApplicationVersions{}, fmt.Errorf("unable to get versions for Marathon app %s", appID)
+	}
+}
+
+func (c *Client) ApplicationByVersion(appID string, version string) (*marathon.Application, error) {
+	dcosClient := pluginutil.HTTPClient(c.baseURL)
```
makkes, line 297: go-marathon has `ApplicationByVersion`. Why not use that?

makkes (changes requested):
CI is also failing

pkg/marathon/client.go
br-lewis: I thought we were mainly use go-marathon for its types rather than the API parts and also, `ApplicationByVersion` works differently (calls `/v2/apps/{}/versions/{}`) which we only want to do in certain circumstances. And this is a more direct porting of the Python code.

makkes (approved):
lgtm, except failing CI

janisz (approved):

PR #409, opened 2019-11-21
Unskip tests

janisz:
After merging we can unskip TLS tests.
* https://github.com/dcos/dcos-cli/pull/1513
* https://github.com/dcos/dcos-cli/pull/1514
* https://github.com/dcos/dcos-core-cli/commit/6e585feb4149d1a803ed0d2d7e45df7d03a04417#r36073375

### DIFF
--- a/python/lib/dcoscli/tests/integrations/test_ssl.py
+++ b/python/lib/dcoscli/tests/integrations/test_ssl.py
@@ -48,7 +48,6 @@ def test_verify_ssl_without_cert_config(env):
         assert stderr.decode('utf-8') == _ssl_error_msg()
 
 
-@pytest.mark.skip(reason="https://jira.mesosphere.com/browse/DCOS-61005")
 def test_verify_ssl_with_bad_cert_env_var(env):
     env['DCOS_SSL_VERIFY'] = 'tests/data/ssl/fake.pem'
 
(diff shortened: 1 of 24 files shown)


python/lib/dcoscli/tests/integrations/test_ssl.py
```diff
@@ -76,5 +74,5 @@ def _ssl_error_msg():
         "run: 'dcos config set core.ssl_verify <value>'\n"
         "<value>: Whether to verify SSL certs for HTTPS or path to certs. "
         "Valid values are a path to a CA_BUNDLE, "
-        "True (will then use CA certificates from certifi), "
+        "True (will then use system CA certificates), "
```
makkes, line 77: Wait, are we testing `dcos/dcos-cli` code here? This is messed up. Anyway, it shouldn't block this PR.

makkes (approved):

python/lib/dcoscli/tests/integrations/test_ssl.py
janisz: Yes we are :( Unfortunately this is the only test to check `ssl_verify` and it uses marathon command.

janisz: OK, it uses marathon command because commands in CLI are not using TLS.

janisz:
:wave: @drewvanstone @br-lewis  @bamarni @rgo3  Can you review, please?

br-lewis (approved):

PR #415, opened 2019-11-25
Fix deprecations

janisz:
In order to fix depredations we need to mark strings as raw (prefix them with `r`).


armandgrillet (approved):

janisz:

makkes (approved):

PR #411, opened 2019-11-22
Change marathon app stop into Go

br-lewis:
Addresses: https://jira.mesosphere.com/browse/DCOS-60277

PR #417, opened 2019-11-25
Port 'dcos marathon app start' in Go.

armandgrillet:
Resolves https://jira.mesosphere.com/browse/DCOS-60276

### DIFF
--- a/pkg/cmd/marathon/app/app_start_test.go
+++ b/pkg/cmd/marathon/app/app_start_test.go
@@ -1,0 +1,101 @@
+package app
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/dcos/dcos-core-cli/pkg/marathon"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMarathonStart(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "/service/marathon/v2/apps/app-test", r.URL.String())
+		switch method := r.Method; method {
+		case http.MethodGet:
+			w.Write([]byte(`{"app":{"id":"/app-test","backoffFactor":1.15,"backoffSeconds":1,"cpus":1.0,"disk":64.0,"executor":"","instances":0}}`))
+		case http.MethodPut:
+			w.Write([]byte(`{"deploymentId":"5ed4c0c5-9ff8-4a6f-a0cd-f57f59a34b43","version":"2015-09-29T15:59:51.164Z"}`))
+		}
+	}))
+	defer ts.Close()
+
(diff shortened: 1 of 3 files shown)


pkg/cmd/marathon/app/app_start_test.go
```diff
@@ -0,0 +1,111 @@
...
+	ctx := mock.NewContext(env)
+	cluster := config.NewCluster(nil)
+	cluster.SetURL(ts.URL)
+	ctx.SetCluster(cluster)
+
+	client, err := marathon.NewClient(ctx)
+	assert.NoError(t, err)
+
+	_, err = marathonAppStart(*client, "app-test", 1, false)
+	assert.EqualError(t, err, "Marathon API error: 404 - Something bad happened!")
```
br-lewis, line 83: In the commands I've ported, I've tried to keep the error messages close to the same as the Python, though I've changed them to meet Go guidelines (no leading uppercase, no period). I'm not too committed to that method but we should pick a consistent way to dealing with errors from `go-marathon`. https://github.com/dcos/dcos-core-cli/pull/411/files#diff-4112de97288590870c674ea78812c862R51-R56

armandgrillet: In the previous commands I've ported I've done the same as you for the most common errors (e.g. 404, here the first 404 has special handling) while letting go-marathon do its things with the others. It seems similar in the snippet you sent.

br-lewis (approved):

pkg/cmd/marathon/app/app_start_test.go
```diff
@@ -0,0 +1,134 @@
...
+	}))
+	defer ts.Close()
+
+	out := new(bytes.Buffer)
+	env := mock.NewEnvironment()
+	env.Out = out
+	ctx := mock.NewContext(env)
+	cluster := config.NewCluster(nil)
+	cluster.SetURL(ts.URL)
+	ctx.SetCluster(cluster)
```
janisz: This code block repeats in every marathon test. Please extract it to a function
```go
func newContext(ts *httptest.Server) (*mock.Context, *bytes.Buffer) {
	out := new(bytes.Buffer)
	env := mock.NewEnvironment()
	env.Out = out
	ctx := mock.NewContext(env)
	cluster := config.NewCluster(nil)
	cluster.SetURL(ts.URL)
	ctx.SetCluster(cluster)
	return ctx, out
}
```

janisz (changes requested):

pkg/cmd/marathon/app/test_util.go
```diff
@@ -0,0 +1,17 @@
+package app
```
janisz: I think we should rename this file from `test_util.go` to `app_test.go` or just put this method into any test file. We shouldl not leak it to a production code since it's only for tests.

janisz (approved):

PR #429, opened 2020-04-01
fix: change aws region

orlandohohmeier:
Change the aws region to resolve cluster creation issues due to configuration issues.


janisz:
> dcos_launch.util.LauncherError: ('ValidationError', "Field: instance_ami, Errors: default value for 'instance_ami' cannot be set: Circular dependencies of default setters.\nrequired field\n")

PR #430, opened 2020-04-14
Use terraform for e2e cluster creation

janisz:
Currently our E2E are broken due to a problem with dcos-launch. 

* switches our E2E tests from dcos-launch to terraform. This will allow us to delete cluster after tests
* adds junit reports for E2E tests
* adjusts small differences in SSH tests output after switch to terraform
* skipped single test needs more investigation [D2IQ-67130](https://jira.d2iq.com/browse/D2IQ-67130)


janisz:
I'll squash all commits on merge

ci/integration-tests-mac.groovy
```diff
@@ -40,25 +40,12 @@ pipeline {
...
-                                        python3 -m venv env; \
-                                        source env/bin/activate; \
-                                        pip -q install --upgrade pip==18.1 setuptools; \
-                                        pip -q install -r requirements.txt; \
-                                        ./launch_aws_cluster.py create"''',
-                                returnStdout: true).trim()
-                    }
-                    stash includes: 'scripts/config.json', name: 'aws-config'
-                }
+//                withCredentials(credentials) {
```
drewvanstone: What's the story here? Is it supposed to be commented out?

drewvanstone (approved):
Looks good. One question about the commented out code.

ci/integration-tests-mac.groovy
janisz: From description:
> disables E2E tests for Mac and Windows. It will be fixed in separated PR

For some reason we get connect timed out for Mac and Windows, for linux too when using py37 docker image. Right now I have no clue why. Can't reproduce it locally :(

pierrebeitz (approved):
learned quite some stuff reviewing this!  
i think a dockerfile for CI would make quite some places less noisy. i also think there are better places to sink time. 😉 

looks good and seems to work. 👍

i'd love to learn what caused the error for the windows- and mac-builds once you found out.

PR #432, opened 2020-04-17
Add junit reports to make test

janisz:
Use junit in jenkins to report tests.


drewvanstone (approved):

ci/integration-tests-linux.groovy
```diff
@@ -33,7 +33,8 @@ pipeline {
     stages {
         stage("Build Go binary") {
             steps {
-                sh "make ${os}"
```
armandgrillet, line 36: This seems to be reverted from what it should be (`make test ${os}`). Is it?

PR #435, opened 2020-04-21
Fix setuptools and PyInstaller

janisz:
Refs: https://github.com/pypa/setuptools/issues/1963#issuecomment-607078946
Fixes: D2IQ-67106


janisz:

PR #445, opened 2020-04-22
Use Jenkins Agent instead of Docker

janisz:
We don't need docker to run on Jenkins since we already have proper images available with label py37.

### DIFF
--- a/ci/build.groovy
+++ b/ci/build.groovy
@@ -31,20 +31,14 @@ pipeline {
         }
         }
         stage('Build Linux binary') {
-          agent {
-            node {
-              label 'mesos'
-              customWorkspace '/workspace'
-            }
-          }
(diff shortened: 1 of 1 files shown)


ci/build.groovy
```diff
@@ -31,20 +31,14 @@ pipeline {
...
+          agent { label 'py37' }
 
           steps {
 
             sh '''
-                docker run --rm -v $PWD:/usr/src -w /usr/src \
-                    python:3.7-stretch bash -exc " \
-                      cd python/lib/dcoscli; \
-                      make binary"
+              bash -exc " \
```
pierrebeitz, line 39: any idea why are we need/want to wrap this in `bash -exc`?

pierrebeitz (approved):

PR #453, opened 2020-06-01
Update dcos node diagnostic deprecation message

janisz:
Backports:  #452
Fixes: COPS-6209

PR #452, opened 2020-06-01
Update dcos node diagnostic deprecation message

janisz:
Fixes: COPS-6209

PR #459, opened 2020-06-02
Bump dependencies

janisz:
Due to https://github.com/dependabot/dependabot-core/issues/670 we need to manually update `vendor`. This PR contains changes from following PRs:
* #455
* #461 
* #462
* #467


janisz:
@mesosphere-ci run Windows integration tests

janisz:
@pierrebeitz Can you take a look?

pierrebeitz:
1. how would i ensure that everything still works?
2. do we trust our tests enough to just merge this?
3. what's the "Revert "Bump github.com/golang/protobuf from 1.3.2 to 1.4.2" about?
4. the description of the PR claims that #455, #453, #452 are contained. i don't see all of them though?

janisz:
**1 & 2** I trust our tests :) We have unit tests and integration tests for Go so it should be safe. Go is statically typed so any breaking changes should be visible during compilation. 

**3.** I need to revert this update because we have problem with protobuf dependency. One of our dependencies (dcos-go) uses protobuf but is not included as go module so we can't update it. I'll fix dcos-go separately. 
**4.**  I fixed the description + I'd like to squash all commits in this PR during merge.

pierrebeitz (approved):

PR #460, opened 2020-06-02
Add JSON output to job run command

janisz:
Add missing --json flag to job run command.
This function was in Python code but was missing in Go.

Fixes: D2IQ-68975

### DIFF
--- a/pkg/cmd/job/job_run.go
+++ b/pkg/cmd/job/job_run.go
@@ -1,6 +1,7 @@
 package job
 
 import (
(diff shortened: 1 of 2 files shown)


pierrebeitz:
how hard would it be to add a test for that?

janisz:
Hmm, adding integration test should be quite easy. I'll do that. Maybe the test existed before.

janisz:

pierrebeitz (approved):
LGTM!

PR #465, opened 2020-06-02
Fix calico test to work with latest version

janisz:
Updated version introduced in https://github.com/dcos/dcos/pull/7272
breaks the test so we need to update desired output.


janisz:
@mesosphere-ci run Windows integration tests

pierrebeitz (approved):

PR #427, opened 2020-03-10
feat(calicoctl): add calicoctl wrapper

Poltergeist:
This adds a wrapper for calicoctl what still needs to be done:
- [x] add calicoctl download on build and bundle it with the plugin
- [x] fix permission passing to calicoctl
- [x] add probing
- [ ] add tests
- [x] add go routines and channels for the requests to improve performance of the calico command

to get this working you currently need to run 
```
curl -L https://github.com/projectcalico/calicoctl/releases/download/v3.12.0/calicoctl-darwin-amd64 --output build/darwin/calico.zip && unzip build/darwin/plugin/bin/calicoctl && chmod +x build/darwin/plugin/bin/calicoctl
```
 before  building the package.

PR #489, opened 2020-08-20
Simplify the calico wrapper

janisz:
* Do not download CA cert – use the one provided by dcos-cli
* Check https for enterprise – this will test if cert is loaded
* Remove unused calico client

Fixes: https://jira.d2iq.com/browse/COPS-6353


pkg/cmd/calico/calico.go
```diff
@@ -99,50 +97,44 @@ func getEnvironment(ctx api.Context, grpcPort string) ([]string, error) {
...
-	caClient := newClient(httpClient)
-	caCert, err := caClient.getCaCertificate()
-	if err != nil {
-		return nil, fmt.Errorf("could not get certificate: %s", err)
+	tlsCAPath, ok := os.LookupEnv("DCOS_TLS_CA_PATH")
+	if !ok {
+		return nil, fmt.Errorf("DCOS_TLS_CA_PATH is not defined. " +
+			"Calico requires secure connection. " +
+			"Do NOT use --insecure flag with dcos cluster setup, " +
+			"--no-check should be used instead")
```
pierrebeitz, line 120: nice 2 know!

pkg/cmd/calico/calico.go
```diff
@@ -99,50 +97,44 @@ func getEnvironment(ctx api.Context, grpcPort string) ([]string, error) {
...
 			fmt.Sprintf("ETCD_CUSTOM_GRPC_METADATA=authorization:token=%s", cluster.ACSToken()),
 			fmt.Sprintf("ETCD_ENDPOINTS=%s%s", host, grpcPort),
 		}, nil
 	}
 
-	caClient := newClient(httpClient)
-	caCert, err := caClient.getCaCertificate()
-	if err != nil {
-		return nil, fmt.Errorf("could not get certificate: %s", err)
+	tlsCAPath, ok := os.LookupEnv("DCOS_TLS_CA_PATH")
```
pierrebeitz, line 115: formerly we requested it via `/ca/dcos-ca.crt`, right? 
why can we change to this method without affecting operators?

janisz: > formerly we requested it via /ca/dcos-ca.crt, right?

Yes and we download this cert every time duplication job that is done during `dcos cluster setup`. [dcos-cli sets this variable](https://github.com/dcos/dcos-cli/blob/b8f810a03d4e989ec02bb98d6685e4eff2d7ceb5/design/plugin.md) for us so we do not need to download cert every time user call `dcos calico`.  

>- `DCOS_TLS_CA_PATH`: Unless `DCOS_TLS_INSECURE=1`, this env var indicates the path to the CA bundle to
                      verify server certificates against. When it is not set, the system's CA bundle must be used.

I mentioned it before https://github.com/dcos/dcos-core-cli/pull/427#discussion_r390294992 but I have no time to properly debug the problem.

pkg/cmd/calico/calico.go
janisz: Refs: https://github.com/dcos/dcos-cli/pull/1535/files

pkg/cmd/calico/calico.go
pierrebeitz: i love this cleanup!

pierrebeitz (approved):

PR #490, opened 2020-08-21
Bump Go to 1.15

janisz:
Currently DC/OS CLI keeps backward compatibility for macOS 10.10 Yosemite (2014) after updating Go version we will drop it and support only macOS 10.12 Sierra or later.

It's now the industry standard to deprecate TLS 1.0 and 1.1.
See: https://tools.ietf.org/html/draft-ietf-tls-oldversions-deprecate-00

Refs: dcos/dcos-cli#1552


pierrebeitz (approved):
LGTM

PR #519, opened 2020-10-28
Fix NaN% in quota list

janisz:
Prevent dividing by 0 in `quota list` command

Fixes: COPS-6618


pierrebeitz (approved):
that looks fantastic!

janisz:
Diff looks messy but I only extracted a function to make testing easier and invert condition to reduce indent level.

pierrebeitz:
yeah, if you diff without whitespace and move some lines, it's not that bad.

